PR #38, opened 2025-11-01
Add get_noise_multiplier API for DP-wrapped models

### LINKED ISSUE #26: Provide a way to get the calculated noise multiplier through public Keras API
Feed back from a user:

"""
It seems like when i don't set a specific noise multiplier in the DP_Keras_Config and instead let the API determine one for the given epsilon budget the field in the original DP_Keras_Config object is no correctly updated.
More concretely when i do:
```
params = keras_api.DPKerasConfig(
                epsilon=FLAGS.epsilon,
                delta=5.1-05,
                clipping_norm=10.0,
                batch_size=1024,
                gradient_accumulation_steps=1,
                train_steps=1900,
                train_size=19601,
                seed=42,
            )
model = keras_api.make_private(model, params)
```
I get the following correct print statement: 

"Finished calculating noise multiplier: calculated_noise_multiplier=0.4813812471830989"

But then:

```
print(params.noise_mutiplier)
```

stil yields "None".
For logging purposes it would be great to expose the DPkeras config with the updated field to the user i think :thinking_face: ...
"""

`params` are not updated because it is immutable object. It should probably stay so. We should provide a way to get the noise_multiplier via public api.

Aaditya93:
Title
Add get_noise_multiplier(model) API to expose calculated noise multiplier (Fixes #26)

Description
Fixes #26 — When DPKerasConfig.noise_multiplier is None, the auto-calculated value was not accessible because the config is immutable. This update introduces a public API to retrieve the effective noise multiplier from a DP-wrapped Keras model.

What Changed
* Added get_noise_multiplier(model: keras.Model) in keras_api.py
    * Returns explicit or auto-calculated value
    * Raises a clear error for non-DP models
* Added precomputation and caching of noise multiplier in make_private() to prevent double-calculation
* Exported get_noise_multiplier in jax_privacy/keras/__init__.py
* Added unit tests for:
    * Explicit noise multiplier retrieval
    * Auto-calculated noise multiplier retrieval
    * Error handling for non-DP models
* Updated examples to demonstrate API usage

Why
Maintains DPKerasConfig immutability while giving users a simple, reliable way to log or inspect the effective noise multiplier for transparency and debugging.

Testing
* New unit tests added and passing locally
* Example confirms noise multiplier is correctly printed before training
* No breaking changes to existing APIs

### DIFF
--- a/jax_privacy/keras/keras_api.py
+++ b/jax_privacy/keras/keras_api.py
@@ -729,3 +729,4 @@ def get_noise_multiplier(model: keras.Model) -> float:
   # make_private ensures that noise_multiplier is always set.
   assert dp_params.noise_multiplier is not None
   return dp_params.noise_multiplier
+  


jax_privacy/keras/keras_api.py
```diff
@@ -240,6 +240,10 @@ def make_private(model: keras.Model, params: DPKerasConfig) -> keras.Model:
   # 4. We replace the model._update_metrics_variables method with a new method
   #    that updates the metrics variables for DP-SGD training.
 
+  # Pre-calculate noise multiplier if not provided
+  if params.noise_multiplier is None:
```
RamSaw, line 244: then we should remove 

```
noise_multiplier = (
      dp_params.noise_multiplier
      if dp_params.noise_multiplier is not None
      else dp_params.update_with_calibrated_noise_multiplier().noise_multiplier
  )
```

in `_noised_clipped_grads` function

jax_privacy/keras/keras_api.py
```diff
@@ -700,3 +704,38 @@ def _get_random_int64() -> np.int64:
...
+  
+  dp_params = model._dp_params  # pylint: disable=protected-access
+  
+  # If noise_multiplier was provided, return it directly
+  if dp_params.noise_multiplier is not None:
+    return dp_params.noise_multiplier
+  
+  # Otherwise, calculate it (this will use the cached value if already computed)
+  updated_params = dp_params.update_with_calibrated_noise_multiplier()
+  return updated_params.noise_multiplier
```
RamSaw: nit: add a new line in the end of the file

jax_privacy/keras/keras_api.py
```diff
@@ -700,3 +704,38 @@ def _get_random_int64() -> np.int64:
...
+  if not hasattr(model, '_dp_params'):
+    raise ValueError(
+        "Model does not appear to be wrapped with make_private(). "
+        "Call make_private() before get_noise_multiplier()."
+    )
+  
+  dp_params = model._dp_params  # pylint: disable=protected-access
+  
+  # If noise_multiplier was provided, return it directly
+  if dp_params.noise_multiplier is not None:
```
RamSaw: we come here only if make_private was called, then it means dp_params must have noise multiplier, so we can just return here asserting that it is not None.

jax_privacy/keras/keras_api.py
```diff
@@ -700,3 +704,38 @@ def _get_random_int64() -> np.int64:
...
+    return dp_params.noise_multiplier
+  
+  # Otherwise, calculate it (this will use the cached value if already computed)
```
RamSaw: just fyi, there is no caching because it is not cached property

jax_privacy/keras/keras_api_test.py
```diff
@@ -476,6 +476,37 @@ def _check_distribution(
...
+
+  def test_get_noise_multiplier_auto_calculated(self):
+    """Test get_noise_multiplier with auto-calculated noise multiplier."""
+    model = keras.Sequential([keras.layers.Dense(10, input_shape=(784,))])
+    params = dataclasses.replace(self._get_params(), noise_multiplier=None)
+    
+    model = keras_api.make_private(model, params)
+    retrieved_noise_mult = keras_api.get_noise_multiplier(model)
+    
+    self.assertIsInstance(retrieved_noise_mult, float)
```
RamSaw: is it really necessary?

RamSaw (changes requested):
Thank you very much for such a great contribution!

In general LGTM but left few minor comments.

In PR description you write that you update examples, but in PR changed files there are no changes to examples. Not saying that it is necessary, without it PR is also fine, but just doulbe checking in case you forgot to push a commit with these changes.

jax_privacy/keras/keras_api_test.py
Aaditya93: Good point — this check is probably redundant since get_noise_multiplier is already type-hinted to return a float.

jax_privacy/keras/keras_api.py
Aaditya93: Ohh got it.

jax_privacy/keras/keras_api.py
Aaditya93: Thanks! Since make_private guarantees that noise_multiplier is initialized, I went ahead and replaced the conditional logic with an assertion to keep things cleaner.

jax_privacy/keras/keras_api.py
Aaditya93: Sorry about that. Fixed

Issue #84, opened 2026-01-06
Use proper Poisson sampling in documented examples and keras_api


ryan112358:
The JAX library core library includes APIs and Implementations of batch selection strategies, including standard Poisson sampling.  However, some of our example binaries do not use this, and instead load in the data sequentially as fixed size batches, which is common practice without privacy.  This is a problematic flaw with the implementations that invalidates the privacy guarantee and should be fixed. 

An example that does correct sampling is given in dp_logistic_regression.py.

Task:
* Update keras_api.py so that Poisson sampling is used instead of deterministic fixed-batch sampling. This may require imposing additional constraints on the input dataset format. In particular, the dataset needs to support efficient random access (in-memory arrays, or arrayrecord format).

PR #89, opened 2026-01-07
bench: refactor `keras_api.py` to use Poisson Sampling 

Neerajpathak07:
Fixes:- #84 

> What is the purpose of this pull request?

This pull request:-

- Refactors `keras_api.py` implementation to use Poisson Sampling from in-house `batch_selection` parity. 
- Updates the data class examples for DP-SGD to a Keras model.
- Removes the previous implementation of deterministic fixed-batch sampling and creates a new function to use Poisson sampling.. 
   Ref:-
   ```python
   return original_fit_fn(
        *args,
        **kwargs,
    )
    ```
   
    With

   ```python
    return fit_with_dp_poisson_sampling
    ```


jax_privacy/keras_api.py
```diff
@@ -335,78 +343,59 @@ def fit_fn_with_validation(
...
+          self._dp_params.sampling_prob or (self._dp_params.batch_size / self._dp_params.train_size)
+      )
+      plan = batch_selection.CyclicPoissonSampling(
+          sampling_prob=sampling_probability,
+          iterations=remaining_steps
+      )
+      for batch_idx in plan.batch_iterator(self._dp_params.train_size):
 
+        # Pad indices to reduce JIT recompilations
+        idx = batch_selection.pad_to_multiple_of(batch_idx, self._dp_params.padding_multiple)
```
ryan112358, line 492: It's important that the padding indices (idx == -1) are plumbed through jax_privacy.clipped_grad, and optionally that a dummy example is appended to the end of the training data that will be picked up when idx == -1

jax_privacy/keras_api.py
```diff
@@ -335,78 +343,59 @@ def fit_fn_with_validation(
...
+    def fit_with_dp_poisson_sampling(x=None, y=None, epochs=1, verbose='auto', callbacks=None, **kwargs):
+      if x is None or not hasattr(x, 'shape'):
+        raise ValueError(
+            'DP training requires x to be a JAX or NumPy array for efficient random access during Poisson sampling.'
+        )
+      # Use provided sampling_prob or default to batch_size / train_size for expected batch size
+      sampling_probability = (
+          self._dp_params.sampling_prob or (self._dp_params.batch_size / self._dp_params.train_size)
+      )
+      plan = batch_selection.CyclicPoissonSampling(
```
ryan112358: maybe call this sampler or strategy

jax_privacy/keras_api.py
```diff
@@ -111,6 +117,8 @@ class DPKerasConfig:
   rescale_to_unit_norm: bool = True
   microbatch_size: int | None = None
   seed: int | None = None
+  sampling_prob: float | None = None
```
ryan112358: I believe exactly one of sampling_prob or batch_size should be provided.  Otherwise another way you could parameterize it is with "expected_batch_size" and a boolean flag "use_poisson_sampling".  Then you can compute sampling_probability below as expected_batch_size / dataset_size.  This would assume a zero-out adjacency notion though.

jax_privacy/keras_api.py
```diff
@@ -470,7 +459,7 @@ def _dp_train_step(
 
   dp_batch_size = self._dp_params.batch_size  # pylint: disable=protected-access
   actual_batch_size = jax.tree_util.tree_leaves(x)[0].shape[0]
-  if dp_batch_size != actual_batch_size:
+  if actual_batch_size < dp_batch_size:
```
ryan112358: Why should this invariant hold?

jax_privacy/keras_api.py
Neerajpathak07:
```suggestion
      strategy= batch_selection.CyclicPoissonSampling(
```

jax_privacy/keras_api.py
Neerajpathak07: @ryan112358 makes sense I'll update this with the suggested parameters. Although if it is going to hold zero-out adjacency notion would changing this be proclamatory?

jax_privacy/keras_api.py
Neerajpathak07: Thought of preserving the DP-SGD privacy assumptions under Poisson sampling. But, `actual_batch_size` may be less than `dp_batch_size` for any given step. So, it would create discrepancy I'll revert this!!

jax_privacy/keras_api.py
```diff
@@ -335,78 +339,82 @@ def fit_fn_with_validation(
...
 
+      clipped_grad_fn = jax_privacy.clipped_grad(
+          fun=self.compute_loss_and_updates,
+          has_aux=True,
+          return_values=True,
+          l2_clip_norm=self._dp_params.clipping_norm,
+          rescale_to_unit_norm=self._dp_params.rescale_to_unit_norm,
+          normalize_by=self._dp_params.batch_size,
+          batch_argnums=(3, 4, 5),  # corresponding to (x, y, sample_weight)
+          microbatch_size=self._dp_params.microbatch_size,
```
ryan112358: If microbatch_size is specified with padding_multiple, then  padding_multiple % microbatch_size == 0 should be true.  Would be good to document that.

jax_privacy/keras_api.py
```diff
@@ -335,78 +339,82 @@ def fit_fn_with_validation(
...
+      privatizer = strategy.noise_addition_transform
+
+      # Append a dummy example for padding
+      dummy_x = jnp.zeros_like(x[:1])
+      x = jnp.concatenate([x, dummy_x], axis=0)
+      if y is not None:
+        dummy_y = jnp.zeros_like(y[:1])
+        y = jnp.concatenate([y, dummy_y], axis=0)
+
+      clipped_grad = clipped_grad_fn(x, y, is_padding_example=is_padding_example)
```
ryan112358: It seems is_padding_example is called here before being used.  The unit tests should catch this

jax_privacy/keras_api.py
```diff
@@ -335,78 +339,82 @@ def fit_fn_with_validation(
...
+      step_logs, updated_model_state = privatizer.update(clipped_grad)
+
+      sampling_probability = self._dp_params.batch_size / self._dp_params.train_size
+
+      strategy = batch_selection.CyclicPoissonSampling(
+          sampling_prob=sampling_probability,
+          iterations=remaining_steps
+      )
+
+      for batch_idx in strategy.batch_iterator(self._dp_params.train_size):
```
ryan112358: I'm confused by this loop.  It seems to iterate through all the batches but not really do anything with the batch_data.

Neerajpathak07: yea was a bug there updated it with:-
```python
step_logs, updated_model_state = self._dp_train_step(
        self,
        model_state,
        batch_data
      )
```

Neerajpathak07:
@ryan112358 Have added a few fixes addressing the comments. But, all-in-all I am having a few doubts on resolving the CI error!!

jax_privacy/keras_api.py
```diff
@@ -335,60 +345,130 @@ def fit_fn_with_validation(
     _validate_optimizer(self, self._dp_params)  # pylint: disable=protected-access
     fit_signature = inspect.signature(original_fit_fn)
 
-    # batch_size is not set explicitely in the fit() call if the input dataset
-    # is already batched. In this case, we assume that the batch sizes are
-    # aligned and use the batch size from the DP parameters. We will check that
-    # the batch sizes are aligned in the train_step function.
+      # batch_size is not set explicitely in the fit() call if the input dataset
```
ryan112358, line 399: Can you revert these formatting changes?

jax_privacy/keras_api.py
```diff
@@ -335,60 +345,130 @@ def fit_fn_with_validation(
...
+            microbatch_size=self._dp_params.microbatch_size,
+        )
+    sampling_probability = self._dp_params.batch_size / self._dp_params.train_size # pylint: disable=protected-access
+
+    strategy = batch_selection.CyclicPoissonSampling( # pylint: disable=protected-access
+            sampling_prob=sampling_probability,
+            iterations=self._dp_params.train_steps - performed_optimizer_steps,
+        )
+  
+    def fit_with_dp_poisson_sampling(*args, x=None, y=None, is_padding_example):
```
ryan112358, line 469: is_padding_example is not used, either here or below.  It should be wired through to clipped_grad_fn

jax_privacy/keras_api.py
```diff
@@ -335,60 +345,130 @@ def fit_fn_with_validation(
...
+        batch_y = y[idx] if y is not None else None
+        batch_data = (batch_x, batch_y, None)
+
+        trainable_variables = self.trainable_variables
+        non_trainable_variables = self.non_trainable_variables
+        optimizer_variables = self.optimizer_variables
+        metrics_variables = self.metrics_variables
+        model_state = (trainable_variables, non_trainable_variables, optimizer_variables, metrics_variables)
+
+        updated_model_state = _dp_train_step(
```
ryan112358, line 504: Maybe I'm missing something, but I don't see where this function is defined

jax_privacy/keras_api.py
```diff
@@ -325,7 +336,6 @@ def _create_fit_fn_with_validation(
     The fit function with same signature as original_fit_fn but with validation
     for DP-SGD training.
   """
-
   @functools.wraps(original_fit_fn)
   def fit_fn_with_validation(
```
ryan112358, line 391: If we make Poisson sampling the default (which requires in-memory arrays as input), we have to make sure that all usages of this API provides the data in the expected format.  Examples that provide data in another format either have to be modified, or Poisson sampling should be optional.  Do you know what format examples of this API use for the data?

jax_privacy/keras_api.py
```diff
@@ -98,6 +99,10 @@ class DPKerasConfig:
         `gradient_accumulation_steps` operates outside of the jit boundary.
         The default value is None, which means that no microbatching is used,
         and is equivalent to microbatch_size=batch_size.
+      padding_multiple: The multiple to pad batches to when using Poisson
+        sampling, to reduce JIT recompilation. Note: If microbatch_size is 
+        specified, then padding_multiple % microbatch_size == 0 should be true
```
RamSaw, line 131: nit: could you write "must" as below you throw an exception if it is not true.

jax_privacy/keras_api.py
RamSaw, line 469: This new function fit_with_dp_poisson_sampling and the associated logic for Poisson sampling appears to be on the right track, but contains several issues and seems to be unused.



The function is defined but never called. The original original_fit_fn is called on L507, so this new logic is currently dead code.

There's a return fit_with_dp_poisson_sampling statement (L505) inside the for loop. This would cause the loop to terminate after the first iteration and return the function object itself, which is likely not the intended behavior.

The is_padding_example parameter in the function signature is shadowed by the assignment on L480.

As noted by the TODO on L462, the clipped_grad_fn is not used.


Note: line numbers might be off

Neerajpathak07: Yeah this is a WIP for now Let me go ahead and draft this!!

Issue #86, opened 2026-01-06
Add simple end-to-end example of DP training of a transformer


ryan112358:
We need a realistic example usage of the library that is simple (ideally 1 file) and easily forkable so that people can easily modify the training loop for research purposes. It should use the Jax privacy core library components rather than the higher level keras_api, the latter is primary for a managed training loop experience, and customization is harder / requires modifications to the Jax privacy core library.

PR #90, opened 2026-01-08
feat: add `examples/dp_sgd_transformer.py` for DP training of a transformer.

Neerajpathak07:
Fixes:- #86

> What is the purpose of this pull request?

This pull request:-

- Adds a single file to `jax_privacy/examples` to add a simple end-to-end and realistic example usage of the library.
- Utilizes core components like, `optax`, `clipped_grad` & `noise_addition`.
- References:- https://github.com/google-deepmind/jax_privacy/blob/main/examples/jax_api_example.py.

### DIFF
--- a/examples/dp_logistic_regression.py
+++ b/examples/dp_logistic_regression.py
@@ -14,6 +14,8 @@
 
 """Trains a logistic regression model with DP-BandMF."""
 
+from typing import Any, Mapping, Tuple
+
 from absl import app
 import jax
 import jax.numpy as jnp
@@ -22,7 +24,6 @@ from jax_privacy import batch_selection
 from jax_privacy.experimental import execution_plan
 import numpy as np
 import optax
-
 
 USERS = 100_000
 FEATURES = 10
@@ -39,14 +40,22 @@ L2_CLIP_NORM = 1.0
 PADDING_MULTIPLE = 32
 
 
-def logistic_loss(params, feature_matrix, labels):
+def logistic_loss(
+    params: Mapping[str, jax.Array],
+    feature_matrix: jax.Array,
+    labels: jax.Array,
+) -> jax.Array:
   logits = jnp.dot(feature_matrix, params['weights']) + params['bias']
   y_pred = 1 / (1 + jnp.exp(-logits))
   y_pred = jnp.clip(y_pred, a_min=1e-6, a_max=1 - 1e-6)
   return -jnp.mean(labels * jnp.log(y_pred) + (1 - labels) * jnp.log1p(-y_pred))
 
 
-def create_benchmark(samples: int, features: int, seed: int = 0):
+def create_benchmark(
+    samples: int,
+    features: int,
+    seed: int = 0,
+) -> Tuple[Mapping[str, jax.Array], jax.Array, jax.Array]:
   """Creates a simple logistic regression model and training data."""
   key = jax.random.key(seed)
   data_key, params_key = jax.random.split(key)
@@ -90,7 +99,13 @@ def main(_):
   privatizer = plan.noise_addition_transform
 
   @jax.jit
-  def update_fn(params, batch, is_padding_example, noise_state, opt_state):
+  def update_fn(
+      params: Mapping[str, jax.Array],
+      batch: Tuple[jax.Array, jax.Array],
+      is_padding_example: jax.Array,
+      noise_state: Any,
+      opt_state: optax.OptState,
+  ) -> Tuple[Mapping[str, jax.Array], Any, optax.OptState]:
     x, y = batch
     clipped_grad = grad_fn(params, x, y, is_padding_example=is_padding_example)
 
--- a/examples/jax_api_example.py
+++ b/examples/jax_api_example.py
@@ -25,6 +25,8 @@ w and b should be very close to the true w and b (max absolute error should be
 smaller than 0.3).
 """
 
+from typing import Any, Mapping, Tuple
+
 from absl import app
 import dp_accounting
 import jax
@@ -38,7 +40,7 @@ from jax_privacy.accounting import calibrate
 import tensorflow as tf
 
 
-def init_model_params():
+def init_model_params() -> Mapping[str, jax.Array]:
   """Initializes the model's weight (w) and bias (b)."""
   key = random.key(12)
   w_key, b_key = random.split(key)
@@ -47,18 +49,23 @@ def init_model_params():
   return {"w": w, "b": b}
 
 
-def model(params, x):
+def model(params: Mapping[str, jax.Array], x: int) -> int:
   """Defines the linear regression model."""
   return params["w"] * x + params["b"]
 
 
-def loss_fn(params, x, y):
+def loss_fn(params: Mapping[str, jax.Array], x: int, y: int) -> jax.Array:
   """Calculates the mean squared error loss."""
   predictions = model(params, x)
   return jnp.mean((predictions - y) ** 2)
 
 
-def load_data(num_samples, true_w=2.0, true_b=1.0, noise_std=0.1):
+def load_data(
+    num_samples: Any,
+    true_w: float = 2,
+    true_b: float = 1,
+    noise_std: float = 0.1,
+) -> Tuple[jax.Array, jax.Array]:
   """Generates synthetic linear regression data."""
   key = random.key(3)
   x = random.uniform(key, (num_samples,), minval=0.0, maxval=10.0)
@@ -67,7 +74,11 @@ def load_data(num_samples, true_w=2.0, true_b=1.0, noise_std=0.1):
   return x, y
 
 
-def batch_dataset(x, y, batch_size):
+def batch_dataset(
+    x: jax.Array,
+    y: jax.Array,
+    batch_size: int,
+) -> tf.data.Dataset:
   """Batches the data into batches of the given size."""
   return (
       tf.data.Dataset.from_tensor_slices((x, y))
@@ -77,13 +88,17 @@ def batch_dataset(x, y, batch_size):
   )
 
 
-def updated_model_params(model_params, noisy_grads):
+def updated_model_params(
+    model_params: Mapping[str, jax.Array],
+    noisy_grads: Mapping[str, jax.Array],
+) -> Mapping[str, jax.Array]:
   """Updates the model parameters using the given gradients."""
   lr = 0.001
   return jax.tree.map(lambda p, g: p - lr * g, model_params, noisy_grads)
 
 
 def main(_):
+
   true_w = 2.0
   true_b = 1.0
   num_epochs = 100
@@ -127,7 +142,12 @@ def main(_):
   noise_state = privatizer.init(model_params)
 
   @jax.jit
-  def dp_update_step(model_params, batch_x, batch_y, noise_state):
+  def dp_update_step(
+      model_params: Mapping[str, jax.Array],
+      batch_x: jax.Array,
+      batch_y: jax.Array,
+      noise_state: Any,
+  ) -> Tuple[Mapping[str, jax.Array], jax.Array, Any]:
     """Updates the model parameters using DP-SGD."""
     grads, aux_outputs = grad_and_value_fn(model_params, batch_x, batch_y)
     loss = aux_outputs.values.mean()
@@ -139,7 +159,11 @@ def main(_):
   # DP only end.
 
   @jax.jit
-  def update_step(model_params, batch_x, batch_y):
+  def update_step(
+      model_params: Mapping[str, jax.Array],
+      batch_x: jax.Array,
+      batch_y: jax.Array,
+  ) -> Tuple[Mapping[str, jax.Array], jax.Array]:
     """Updates the model parameters without DP."""
     loss, grads = jax.value_and_grad(loss_fn, has_aux=False)(
         model_params, batch_x, batch_y
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,8 @@ test = [
 examples = [
     "dinosaur",
     "equinox",
-    "orbax",
+    # orbax pulls orbax-checkpoint → uvloop (not supported on Windows)
+    "orbax; sys_platform != 'win32'",
     "scikit-learn==1.5.0",
     "grain",
     "keras",


examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,293 @@
...
+  logits = model(model_params, batch_x)
+  logits = logits[:, :-1, :].reshape(-1, logits.shape[-1])
+  targets = batch_y[:, 1:].reshape(-1)
+  return optax.softmax_cross_entropy_with_integer_labels(logits, targets).mean()
+
+
+def load_text_data(max_len=128):
+  """Loads and preprocesses text data."""
+  # Use a small text dataset
+  text = "Hello world! This is a simple example of text for training a transformer with differential privacy. " * 1000
```
ryan112358: Can you use a standard text dataset like IMDB or tiny Shakespeare?

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,293 @@
...
+  seqs = []
+  for i in range(0, len(data) - max_len - 1, max_len // 2):
+    seq = data[i:i+max_len+1]  # +1 for target
+    seqs.append(seq)
+
+  seqs = jnp.array(seqs)
+  return seqs, vocab_size, char_to_idx, idx_to_char
+
+
+def batch_dataset(data, batch_size, shuffle=True):
```
ryan112358: Since the data fully fits into memory, it would be good to use jax_privacy.batch_selection

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,293 @@
...
+  if shuffle:
+    dataset = dataset.shuffle(buffer_size=len(data))
+  dataset = dataset.batch(batch_size, drop_remainder=True)
+  return dataset
+
+
+def update_model_params(model_params, noisy_grads):
+  """Updates parameters using gradients."""
+  lr=0.001
+  return jax.tree.map(lambda p, g: p - lr * g, model_params, noisy_grads)
```
ryan112358: If you are using optax, I think optax.apply_updates would be better here

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,293 @@
...
+    )
+    noise_state = privatizer.init(model_params)
+
+    @jax.jit
+    def dp_train_step(model_params, batch_data, noise_state):
+      batch_x = batch_data[:, :-1]
+      batch_y = batch_data[:, 1:]
+      grads, aux = grad_fn(model_params, batch_x, batch_y)
+      loss = aux.values.mean()
+      mean_grads = jax.tree.map(lambda g: g / batch_size, grads)
```
ryan112358: sensitivity above is calculated assuming you are comuting a sum-of-grads, but here you are computing a mean.  If you pass in "normalize_by=batch_size" to clipped_grad, you'll get the correct sensitivity

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,293 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Example of training a Transformer on text using JAX Privacy core API.
+
+This example demonstrates how to train a simple Transformer decoder
+on character-level language modeling using differentially private stochastic
+gradient descent (DP-SGD) with the JAX Privacy core library components. It shows how to:
```
ryan112358: Delete "It shows how to:" or complete the thought

ryan112358 (changes requested):

examples/dp_sgd_transformer_example.py
Neerajpathak07: Sure will test it on these!!!

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,301 @@
...
+  logits = logits[:, :-1, :].reshape(-1, logits.shape[-1])
+  targets = batch_y[:, 1:].reshape(-1)
+  return optax.softmax_cross_entropy_with_integer_labels(logits, targets).mean()
+
+
+def load_text_data(max_len=128):
+  """Loads and preprocesses text data."""
+  # Use tiny Shakespeare dataset
+  ds = tfds.load('tiny_shakespeare', split='train')
+  text = ''.join([example['text'].numpy().decode('utf-8') for example in ds.take(100)])
```
Neerajpathak07: @ryan112358 Used tiny Shakespeare!!

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,302 @@
...
+  var = jnp.var(batch_x, axis=-1, keepdims=True)
+  return model_params['scale'] * (batch_x - mean) / jnp.sqrt(var + 1e-5) + model_params['bias']
+
+
+def model(model_params, batch_x):
+  """Transformer forward pass."""
+  seq_len = batch_x.shape[1]
+  batch_x = model_params['embedding'][batch_x] + model_params['pos_embedding'][:seq_len]
+
+  mask = jnp.tril(jnp.ones((seq_len, seq_len)))
```
ryan112358: nit: jnp.tri(seq_len) is simpler

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,302 @@
...
+    )
+    privatizer = noise_addition.gaussian_privatizer(
+        stddev=noise_multiplier * sensitivity, prng_key=noise_rng
+    )
+
+    optimizer = optax.sgd(learning_rate)
+    privatizer = plan.noise_addition_transform
+
+  @jax.jit
+  def dp_train_step(model_params, opt_state, batch_data, noise_state):
```
ryan112358: There seems to be some inconsitencies with the function signature and how it is used below.  Specifically w.r.t is_padding_example and opt_state.

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,302 @@
...
+    batch_x = batch_data[:, :-1]
+    batch_y = batch_data[:, 1:]
+    grads, aux = grad_fn(model_params, batch_x, batch_y)
```
ryan112358: Add is_padding_example here

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,302 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Example of training a Transformer on text using JAX Privacy core API.
+
+This example demonstrates how to train a simple Transformer decoder
+on character-level language modeling using differentially private stochastic
+gradient descent (DP-SGD) with the JAX Privacy core library components.
```
ryan112358: Can you add some description of some of the design choices made in the header like the dataset used, tokenizer (character level), unit of privacy (sequence level), model characteristics like learned positional embedding, etc.

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,302 @@
...
+def transformer_block(model_params, batch_x, mask):
+  """Single transformer block."""
+  # Multi-head attention
+  q = jnp.dot(batch_x, model_params['attn']['q'])
+  k = jnp.dot(batch_x, model_params['attn']['k'])
+  v = jnp.dot(batch_x, model_params['attn']['v'])
+
+  num_heads = 4
+  head_dim = q.shape[-1] // num_heads
+  q = q.reshape(q.shape[0], q.shape[1], num_heads, head_dim).transpose(0, 2, 1, 3)
```
ryan112358: I would suggest either using Jax.nn.dot_product_attention here or using einsum so it's easier to understand

examples/dp_sgd_transformer.py
Neerajpathak07: yeah makes sense, let me streamline this!!

Neerajpathak07:
@ryan112358 Glad to hear you liked it, a big thanks for speed running this PR by providing crucial reviews!!

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,383 @@
...
+  keys = random.split(key, 5)
+  model_params = {
+      'embedding': random.normal(keys[0], (vocab_size, embed_dim)) * 0.1,
+      'pos_embedding': random.normal(keys[1], (max_len, embed_dim)) * 0.1,
+      'layers': [],
+      'ln_f': {'scale': jnp.ones(embed_dim), 'bias': jnp.zeros(embed_dim)},
+      'head': random.normal(keys[4], (embed_dim, vocab_size)) * 0.1,
+  }
+  for _ in range(num_layers):
+    layer_key = random.split(keys[2], 8)
```
RamSaw: we probably want different keys for each layer, we have to call `random.split(keys[2], num_layers * 8)` before the loop and then slice appropriately.

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,383 @@
...
+from jax_privacy.experimental import execution_plan
+
+import optax
+import tensorflow as tf
+import tensorflow_datasets as tfds
+
+
+def init_model_params(
+    key,
+    vocab_size=256,
```
RamSaw: small comment:

This function has several arguments that could be confused if passed positionally. Consider making the optional arguments keyword-only by adding a * to the signature: def init_model_params(key, *, vocab_size=256, ...).

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,383 @@
...
+from jax_privacy.accounting import analysis
+from jax_privacy.accounting import calibrate
+from jax_privacy.experimental import execution_plan
+
+import optax
+import tensorflow as tf
+import tensorflow_datasets as tfds
+
+
+def init_model_params(
```
RamSaw, line 55: could you please add type hints whenever possible? same for other functions

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,383 @@
...
+    model_params = optax.apply_updates(model_params, updates)
+    return model_params, opt_state, loss, noise_state
+
+  noise_state = privatizer.init(model_params)
+  opt_state = optimizer.init(model_params)
+
+  print(f'Training Transformer with {'DP' if use_dp else 'no DP'}...')
+
+  for step, batch_idx in enumerate(
+      plan.batch_selection_strategy.batch_iterator(train_size)
```
RamSaw, line 299: plan won't be defined if `use_dp` is False. I am fine if we remove support for non-dp run at all and just have only dp run.

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,383 @@
...
+  seqs = []
+  for i in range(0, len(data) - max_len - 1, max_len // 2):
+    seq = data[i:i+max_len+1]  # +1 for target
+    seqs.append(seq)
+
+  seqs = jnp.array(seqs)
+  return seqs, vocab_size, char_to_idx, idx_to_char
+
+
+def batch_dataset(data, batch_size, shuffle=True):
```
RamSaw: this function is not used

`update_model_params` below too

examples/dp_sgd_transformer.py
RamSaw: I have some generated suggestions for you, feel free to edit if they are incorrect:
```
from typing import Any, Mapping, Sequence, Tuple

def init_model_params(
    key: jax.Array,
    vocab_size: int = 256,
    embed_dim: int = 128,
    num_layers: int = 2,
    max_len: int = 128
) -> Mapping[str, Any]:

def transformer_block(
    model_params: Mapping[str, Any], batch_x: jax.Array, mask: jax.Array
) -> jax.Array:

def layer_norm(
    batch_x: jax.Array, model_params: Mapping[str, Any]
) -> jax.Array:

def model(model_params: Mapping[str, Any], batch_x: jax.Array) -> jax.Array:

def loss_fn(
    model_params: Mapping[str, Any], batch_x: jax.Array, batch_y: jax.Array
) -> jax.Array:

def load_text_data(
    max_len: int = 128,
) -> Tuple[jax.Array, int, Mapping[str, int], Mapping[int, str]]:

def batch_dataset(
    data: jax.Array, batch_size: int, shuffle: bool = True
) -> tf.data.Dataset:

def update_model_params(
    model_params: Mapping[str, Any], updates: Any
) -> Mapping[str, Any]:

def generate_text(
    model_params: Mapping[str, Any],
    seed_text: str,
    length: int = 50,
    char_to_idx: Mapping[str, int] | None = None,
    idx_to_char: Mapping[int, str] | None = None
  ) -> str:

def main(argv: Sequence[str]) -> None:
  if len(argv) > 1:
    raise app.UsageError('Too many command-line arguments.')

 def dp_train_step(
      model_params: Mapping[str, Any],
      batch_data: Tuple[jax.Array, jax.Array],
      is_padding_example: jax.Array,
      noise_state: Any,
      opt_state: Any
  ) -> Tuple[Mapping[str, Any], Any, jax.Array, Any]:
```

Neerajpathak07: @RamSaw I love these, it actually makes more sense to use `jax_array` and `Tuple` to represent `batch_data`.
We can also do something similar for the other examples listed in `examples/` ig this will make them more readable.

examples/dp_sgd_transformer.py
Neerajpathak07: Yeah I'm thinking of removing these reason being 2 folds:-
- Since `batch_dataset` conditions this for `shuffle` logic this will cause discrepancy for Poisson Sampling.
- As we are using:-
```python
model_params = optax.apply_updates(model_params, updates)
```
explicitly inside:- 
```python 
def dp_train_step()
```

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,380 @@
...
+  model_params = {
+      'embedding': random.normal(keys[0], (vocab_size, embed_dim)) * 0.1,
+      'pos_embedding': random.normal(keys[1], (max_len, embed_dim)) * 0.1,
+      'layers': [],
+      'ln_f': {'scale': jnp.ones(embed_dim), 'bias': jnp.zeros(embed_dim)},
+      'head': random.normal(keys[4], (embed_dim, vocab_size)) * 0.1,
+  }
+  layer_keys = random.split(keys[2], num_layers * 8)
+  for i in range(num_layers):
+    layer_key = layer_keys[i*8:(i+1)*8]
```
Neerajpathak07: Here for every `layer` starting from 0 `layer_keys` used will be:-
- 0 -> `layer_keys`[0:8]
- 1 - > `layer_keys`[8:16]
and so on .
This way it can be sliced accordingly for 8 layers.

@RamSaw LMK if I am thinking straight

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,379 @@
...
+from jax_privacy import batch_selection
+from jax_privacy import noise_addition
+from jax_privacy.accounting import accountants
+from jax_privacy.accounting import analysis
+from jax_privacy.accounting import calibrate
+from jax_privacy.experimental import execution_plan
+from typing import Any, Mapping, Sequence, Tuple
+
+import optax
+import tensorflow_datasets as tfds # pytype: disable=import-error
```
Neerajpathak07: have disabled the `pytype` import error for now. Given that it is installed in py env.
```bash
PS C:\Users\HP\jax-privacy\jax_privacy> pip install tensorflow-datasets
Defaulting to user installation because normal site-packages is not writeable
Requirement already satisfied: tensorflow-datasets in c:\users\hp\appdata\roaming\python\python313\site-packages (4.9.9)
```
Wondering if there is something to do with `Any` type annotation.
Ref:- https://google.github.io/pytype/errors.html#import-error

RamSaw: Could you share the error you get? I would try to avoid disabling it if we can.

examples/dp_sgd_transformer.py
RamSaw: Thank you for adding types here! For other examples I don't mind, but let's make it in a separate PR.

examples/dp_sgd_transformer.py
RamSaw: yeah, seems correct

examples/dp_sgd_transformer.py
```diff
@@ -274,59 +272,57 @@ def main(_):
...
-    )
-    noise_multiplier = calibrate.calibrate_noise_multiplier(
-        target_epsilon=epsilon,
-        target_delta=delta,
-        accountant=accountant,
-        batch_sizes=batch_size,
-        num_updates=num_epochs * (train_size // batch_size),
-        num_samples=train_size,
-    )
+  # Set up DP components
```
RamSaw, line 254: This block of code for DP accounting and noise calibration seems to be unused. The privatizer is created at line 310 but is immediately overwritten at line 315 by plan.noise_addition_transform. So you can drop the accountant, noise_multiplier, noise_rng and first assignment to privatizer. The right way is to use the plan.

examples/dp_sgd_transformer.py
```diff
@@ -274,59 +272,57 @@ def main(_):
...
-        sampling_prob=expected_batch_size / train_size,
-    )
-    plan = config.make(grad_fn)
+  grad_fn = jax_privacy.clipped_grad(
+      loss_fn,
+      l2_clip_norm=clipping_norm,
+      batch_argnums=(1, 2),
+      has_aux=False,
+      return_values=True,
+      normalize_by=batch_size,
```
RamSaw: here we should divide not by batch_size but by expected batch size (`sampling_probability * dataset_size`). Another option is to drop it at all and adjust learning rate

examples/dp_sgd_transformer.py
Neerajpathak07: Ref:- https://github.com/google-deepmind/jax_privacy/actions/runs/21294668461

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,351 @@
...
+  return logits
+
+
+def loss_fn(
+    model_params: Mapping[str, Any], batch_x: jax.Array, batch_y: jax.Array
+) -> jax.Array:
+  """Cross-entropy loss for next token prediction."""
+  logits = model(model_params, batch_x)
+  logits = logits[:, :-1, :].reshape(-1, logits.shape[-1])
+  targets = batch_y[:, 1:].reshape(-1)
```
RamSaw: There seems to be an off-by-one error in the loss calculation.

Here's my understanding of the data flow:
1.  In `main`, `batch_x` is `data[:, :-1]` and `batch_y` is `data[:, 1:]`. This correctly sets up `batch_y` as the target sequence for `batch_x`. For any index `i`, `batch_y[:, i]` is the token that should follow `batch_x[:, i]`.
2.  The model output `logits[:, i]` is the prediction for the token following `batch_x[:, i]`. So, `logits[:, i]` should be compared against `batch_y[:, i]`.
3.  However, here you slice both `logits` and `targets`. `logits[:, :-1, :]` drops the predictions for the last input token, and `batch_y[:, 1:]` drops the first target token.
4.  This results in comparing the model's prediction for token `i+1` (from `logits[:, i]`) with the target for token `i+2` (from `batch_y[:, i+1]`).

I believe you can remove the slicing on both `logits` and `targets` and simply reshape them. This would compute the loss over the entire sequence.

```python
  logits = logits.reshape(-1, logits.shape[-1])
  targets = batch_y.reshape(-1)
```

Could you please double-check if my understanding is correct?

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,351 @@
...
+  batch_x = jnp.array([[char_to_idx.get(c, 0) for c in seed_text]])
+  generated = seed_text
+
+  for _ in range(length):
+    logits = model(model_params, batch_x)
+    next_token = jnp.argmax(logits[0, -1, :])
+    next_char = idx_to_char.get(int(next_token), '?')
+    generated += next_char
+    batch_x = jnp.concatenate([batch_x, jnp.array([[next_token]])], axis=1)
+    if batch_x.shape[1] >= 128:  # max_len
```
RamSaw: This 128 seems to be a hardcoded value for max_len. It would be better to pass max_len as an argument to this function to avoid it getting out of sync with the max_len used for training.

examples/dp_sgd_transformer.py
```diff
@@ -0,0 +1,351 @@
...
+  """Initializes Transformer parameters."""
+  keys = random.split(key, 5)
+  model_params = {
+      'embedding': random.normal(keys[0], (vocab_size, embed_dim)) * 0.1,
+      'pos_embedding': random.normal(keys[1], (max_len, embed_dim)) * 0.1,
+      'layers': [],
+      'ln_f': {'scale': jnp.ones(embed_dim), 'bias': jnp.zeros(embed_dim)},
+      'head': random.normal(keys[4], (embed_dim, vocab_size)) * 0.1,
+  }
+  layer_keys = random.split(keys[2], num_layers * 8)
```
RamSaw: sorry missed it, should it be 6 not 8, or?

examples/dp_sgd_transformer.py
RamSaw: Let's add tensorflow_datasets at this line: https://github.com/google-deepmind/jax_privacy/blob/6983469263848094b493cfbb3b576579f8e09433/pyproject.toml#L50. It should fix it then.

RamSaw: let's add `tesnorflow_datasets` at this line: https://github.com/google-deepmind/jax_privacy/blob/6983469263848094b493cfbb3b576579f8e09433/pyproject.toml#L50

it should solve the issue

RamSaw (changes requested):

RamSaw:
There are some complains from internal linter, I will figure it out too soon.

examples/dp_sgd_transformer.py
Neerajpathak07: You’re absolutely right, thanks for flagging this.
Since `batch_x = data[:, :-1]` and `batch_y = data[:, 1:]` are already aligned for next-token prediction, the additional slicing on both logits and `batch_y` introduces an off-by-one error.

RamSaw (approved):

PR #91, opened 2026-01-09
Add simple end-to-end example of DP training of a transformer

moksha-hub:
This addresses issue #86 by adding a single-file, forkable example that uses JAX Privacy core library (not Keras API), implements a minimal transformer for next-token prediction, demonstrates per-example gradient clipping, noise calibration from privacy budget, and gaussian_privatizer for noise addition.

### DIFF
--- a/examples/dp_sgd_transformer_example.py
+++ b/examples/dp_sgd_transformer_example.py
@@ -36,9 +36,9 @@ import dp_accounting
 import jax
 from jax import random
 import jax.numpy as jnp
+import numpy as np
 import optax
 import jax_privacy
-from jax_privacy import batch_selection
 from jax_privacy import noise_addition
 from jax_privacy.accounting import accountants
 from jax_privacy.accounting import analysis
@@ -222,16 +222,19 @@ def loss_fn(params, input_ids, target_ids):
   return -jnp.mean(target_log_probs)
 
 
-def generate_synthetic_data(key, num_samples):
-  """Generate synthetic next-token prediction data.
+def generate_synthetic_data(seed, num_samples):
+  """Generate synthetic next-token prediction data using numpy.
+  
+  Uses numpy instead of JAX to ensure the full dataset lives on CPU,
+  with only batches transferred to GPU during training.
   
   Generates sequences where each token is predictable from the previous ones
   (simple pattern: each token is (previous_token + 1) mod vocab_size).
   """
-  keys = random.split(key, num_samples)
+  rng = np.random.default_rng(seed)
   
   # Start with random tokens
-  start_tokens = random.randint(keys[0], (num_samples, 1), 0, VOCAB_SIZE)
+  start_tokens = rng.integers(0, VOCAB_SIZE, size=(num_samples, 1))
   
   # Generate sequences with a simple pattern
   sequences = [start_tokens]
@@ -239,7 +242,7 @@ def generate_synthetic_data(key, num_samples):
     next_token = (sequences[-1] + 1) % VOCAB_SIZE
     sequences.append(next_token)
   
-  full_sequences = jnp.concatenate(sequences, axis=1)
+  full_sequences = np.concatenate(sequences, axis=1)
   
   # Input is tokens 0 to seq_len-1, target is tokens 1 to seq_len
   input_ids = full_sequences[:, :-1]
@@ -265,9 +268,9 @@ def main(_):
   key = random.key(42)
   key, init_key, data_key = random.split(key, 3)
   
-  # Generate synthetic data
+  # Generate synthetic data (uses numpy to keep data on CPU)
   print("\nGenerating synthetic data...")
-  input_ids, target_ids = generate_synthetic_data(data_key, train_size)
+  input_ids, target_ids = generate_synthetic_data(seed=42, num_samples=train_size)
   print(f"  Training samples: {train_size}")
   print(f"  Sequence length: {SEQ_LEN}")
   print(f"  Vocabulary size: {VOCAB_SIZE}")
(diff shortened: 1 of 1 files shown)


moksha-hub:
@googlebot I signed it!

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,399 @@
...
+# Transformer hyperparameters
+VOCAB_SIZE = 100
+SEQ_LEN = 16
+D_MODEL = 32
+NUM_HEADS = 2
+D_FF = 64
+NUM_LAYERS = 2
+
+
+def init_transformer_params(key):
```
ryan112358, line 66: You might want to add a @jax.jit annotation here

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,399 @@
...
+
+
+def layer_norm(x, scale, bias, eps=1e-5):
+  """Apply layer normalization."""
+  mean = jnp.mean(x, axis=-1, keepdims=True)
+  var = jnp.var(x, axis=-1, keepdims=True)
+  return scale * (x - mean) / jnp.sqrt(var + eps) + bias
+
+
+def attention(x, wq, wk, wv, wo):
```
ryan112358, line 135: Maybe consider implementing this in terms of Jax.nn.dot_product_attention

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,399 @@
...
+    # [START DP SETUP]
+    print("\nSetting up differential privacy...")
+    
+    # Calculate noise multiplier from privacy budget
+    accountant = analysis.DpsgdTrainingAccountant(
+        dp_accountant_config=accountants.PldAccountantConfig()
+    )
+    num_updates = num_epochs * (train_size // batch_size)
+    
+    noise_multiplier = calibrate.calibrate_noise_multiplier(
```
ryan112358, line 301: Add a comment that this API calibrates noise to the mean gradient rather than the sum of gradients.

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,399 @@
...
+      loss = aux_outputs.values.mean()
+      
+      # Average gradients over batch
+      mean_grads = jax.tree.map(lambda x: x / batch_size, grads)
+      
+      # Add calibrated noise
+      noisy_grads, noise_state = privatizer.update(mean_grads, noise_state)
+      
+      # SGD update
+      params = jax.tree.map(
```
ryan112358: Maybe write this in terms of an optax.GradientTransformation (default to adamw), then use optax.apply_upates here

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,399 @@
...
+          lambda p, g: p - learning_rate * g, params, grads
+      )
+      return params, loss
+  
+  # Training loop
+  print(f"\n{'=' * 60}")
+  print(f"Training {'with DP' if use_dp else 'without DP'}...")
+  print(f"{'=' * 60}")
+  
+  for epoch in range(num_epochs):
```
ryan112358, line 372: Can you use a jax_privacy.batch_selection here?  The accounting assumes Poisson sampling, not shuffling

moksha-hub:
@ryan112358 now can you verify again Thanks for previous reply.

examples/dp_sgd_transformer_example.py
```diff
@@ -357,7 +367,11 @@ def train_step(params, batch_input, batch_target):
   print(f"{'=' * 60}")
   
   for epoch in range(num_epochs):
-    # Shuffle data
+    # Note: For strict privacy accounting, you should use Poisson sampling
+    # (each example included independently with probability batch_size/train_size)
+    # rather than shuffling. For simplicity, we use shuffling here, but in
+    # production you should use jax_privacy.batch_selection for proper sampling.
+    # See: https://github.com/google-deepmind/jax_privacy for batch_selection API.
```
ryan112358: In this case since the inputs are just arrays, you should be able to index into them with Poisson sampled indices without issue, so might as well do that here

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,413 @@
...
+  # Gather log probs for target tokens
+  batch_size, seq_len = target_ids.shape
+  target_log_probs = jnp.take_along_axis(
+      log_probs, target_ids[:, :, None], axis=-1
+  ).squeeze(-1)
+  
+  return -jnp.mean(target_log_probs)
+
+
+def generate_synthetic_data(key, num_samples):
```
ryan112358: Is there any performance difference between using numpy vs. Jax for this function?  Numpy might be preferable here as a way to ensure that the full dataset lives on CPU, and only batches get transferred to GPU via Jax.  Your train_step should handle this automatically

examples/dp_sgd_transformer_example.py
```diff
@@ -0,0 +1,413 @@
...
+      updates, opt_state = optimizer.update(noisy_grads, opt_state, params)
+      params = optax.apply_updates(params, updates)
+      return params, opt_state, loss, noise_state
+    
+  else:
+    @jax.jit
+    def train_step(params, batch_input, batch_target):
+      """Single non-private training step."""
+      loss, grads = jax.value_and_grad(loss_fn)(params, batch_input, batch_target)
+      params = jax.tree.map(
```
ryan112358, line 362: This should also use optax

examples/dp_sgd_transformer_example.py
ryan112358: Please resolve.  You can use jax_privacy.batch_selection.CyclicPoissonSampling to generate batch_idx in a DP-compliant manner.

PR #94, opened 2026-01-12
keras_api: expose noise multiplier getter

### LINKED ISSUE #26: Provide a way to get the calculated noise multiplier through public Keras API

staticpayload:
## Summary
- expose a public helper and model method to read the configured or calibrated noise multiplier
- cache the calibrated value on the model and reuse it in DP-SGD gradient noise
- add tests for config-based and calibrated retrieval

## Testing
- pytest tests/keras_api_test.py (Amazon Linux 2023, Python 3.11, eu-central-2)

Fixes #26

### DIFF
--- a/jax_privacy/keras_api.py
+++ b/jax_privacy/keras_api.py
@@ -1,4 +1,5 @@
 # coding=utf-8
+# pylint: disable=todo-style
 # Copyright 2025 DeepMind Technologies Limited.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");


jax_privacy/keras_api.py
```diff
@@ -256,6 +257,41 @@ def make_private(model: keras.Model, params: DPKerasConfig) -> keras.Model:
...
 
+def get_noise_multiplier(model: keras.Model) -> float:
+  """Returns the noise multiplier used for DP-SGD training.
+
+  If the noise multiplier is not set in DPKerasConfig, this will calibrate it
+  once and cache the value on the model.
+
+  Example:
+    >>> import keras  # doctest: +SKIP
+    >>> from jax_privacy import keras_api  # doctest: +SKIP
```
ryan112358: Why are these doctests skipped?

I'm not sure this doctest belongs on this method,  I think it better fits in the module-level docstring

ryan112358 (changes requested):

jax_privacy/keras_api.py
```diff
@@ -256,6 +281,22 @@ def make_private(model: keras.Model, params: DPKerasConfig) -> keras.Model:
   return model
 
 
+def get_noise_multiplier(model: keras.Model) -> float:
+  """Returns the noise multiplier used for DP-SGD training.
+
+  If the noise multiplier is not set in DPKerasConfig, this will calibrate it
```
ryan112358, line 292: pylint is complaining about the docstring format.  Valid formats include:

a) one-lin edocstring (delete this extra stuff)
b) full docstring, but it needs to include an "Arguments" and "Returns" block

jax_privacy/keras_api.py
```diff
@@ -533,11 +576,29 @@ def _dp_train_step(
...
+    model._dp_noise_multiplier = noise_multiplier  # pylint: disable=protected-access
+  return noise_multiplier
+
+
 def _noised_clipped_grads(
     compute_loss_and_updates_fn: LossFn,
     dp_params: DPKerasConfig,
     state: _StateType,
     data: _KerasInputsDataType,
+    model: keras.Model | None = None,
```
ryan112358, line 621: Add to docstring

ryan112358: I think this is the main thing blocking submission.  Here is the error I see:

Args section is missing, does not match arg list or is badly indented.

Expected: compute_loss_and_updates_fn, dp_params, state, data, model

Found   : compute_loss_and_updates_fn, dp_params, state, data

Missing: model

ryan112358 (approved):

Issue #85, opened 2026-01-06
Add support  and examples for cryptographically secure Jax privacy programs


ryan112358:
DP mechanisms typically have randomness in two places: batch selection and noise addition.  Some programs that use Jax privacy may also require internal randomness, although this randomness does not typically affect the privacy guarantee.  By default Jax privacy uses numpy pseudo-random number generation for batch selection, and Jax pseudo-random number generation with explicit key management. 

The `randomgen` library can be used for generating crypographically secure random values, and conforms to the same API as numpy.  This can offer a drop-in replacement for batch_selection and used via dependency injection by Jax privacy users.  For noise_addition, this library can be used to generate normal random values in numpy (on CPU).  For efficiency, this should be done asyncronously with the calculation + accumulation of per-example clipped gradients (which occurs on GPU/TPU).  

Task: 
* Add a new example binary that demonstrates DP training with cryptographically secure randomness, adding useful reusable components to the Jax privacy library as necessary/helpful.

PR #100, opened 2026-01-17
FIX: Add cryptographically secure batch selection and DP training example using randomgen

amyssnippet:
FIXED THE ISSUE #85 

- Introduces CryptoSecureBatchSelectionStrategy in the Jax Privacy library, enabling batch selection with cryptographically secure randomness via the randomgen library.
- Refactors batch selection internals to support dependency injection of secure RNGs.
- Adds a new example binary (dp_training_crypto_secure.py) demonstrating differentially private training with cryptographically secure randomness for both batch selection and noise addition.
- Ensures extensibility for users requiring secure randomness in privacy-sensitive applications.

### DIFF
--- a/examples/dp_training_crypto_secure.py
+++ b/examples/dp_training_crypto_secure.py
@@ -1,0 +1,48 @@
+"""
+Example: DP training with cryptographically secure randomness using Jax Privacy and randomgen.
+"""
+import numpy as np
+try:
+    import randomgen
+except ImportError:
+    raise ImportError("randomgen must be installed for this example.")
+
+from jax_privacy.batch_selection import CryptoSecureBatchSelectionStrategy
+from jax_privacy.noise_addition import gaussian_privatizer
+import jax
+import jax.numpy as jnp
+import optax
+
+# Dummy dataset and model for demonstration
+NUM_EXAMPLES = 1000
+BATCH_SIZE = 128
+EPOCHS = 2
+
+# Create a cryptographically secure RNG
+crypto_rng = randomgen.RandomGenerator(randomgen.Xoshiro256(secure=True))
+
+# Batch selection strategy using cryptographically secure RNG
+batch_selector = CryptoSecureBatchSelectionStrategy()
+batch_iter = batch_selector.batch_iterator(NUM_EXAMPLES, rng=crypto_rng)
+
+# Dummy model parameters and gradients
+params = jnp.zeros((10,))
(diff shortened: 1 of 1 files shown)


jax_privacy/batch_selection.py
```diff
@@ -52,25 +52,53 @@ class PartitionType(enum.Enum):
 def _independent_partition(
     num_examples: int,
     num_groups: int,
-    rng: np.random.Generator,
+    rng,
     dtype: np.typing.DTypeLike
-) -> list[np.ndarray]:
-  sizes = rng.multinomial(num_examples, np.ones(num_groups) / num_groups)
-  boundaries = np.cumsum(sizes)[:-1]
-  indices = np.random.permutation(num_examples).astype(dtype)
```
ryan112358, line 60: I thin this should be rng.permutation actually, the usage of np.random is a mistake.  This will allow you to delete the if / else as wella s the local import of randomgen

jax_privacy/batch_selection.py
```diff
@@ -52,25 +52,53 @@ class PartitionType(enum.Enum):
...
-  return [g[:group_size] for g in groups]
+  ) -> list[np.ndarray]:
+    try:
+      import randomgen
+    except ImportError:
+      randomgen = None
+    if randomgen and isinstance(rng, randomgen.RandomGenerator):
+      indices = rng.permutation(num_examples).astype(dtype)
+    else:
+      indices = np.random.permutation(num_examples).astype(dtype)
```
ryan112358, line 86: Same here

jax_privacy/batch_selection.py
ryan112358, line 1: Changes to this file should not be necessary (except the bug fix of np.random -> rng).  You should be able to just pass in a secure rng via dependency injection

examples/dp_training_crypto_secure.py
```diff
@@ -0,0 +1,48 @@
+"""
+Example: DP training with cryptographically secure randomness using Jax Privacy and randomgen.
+"""
+import numpy as np
+try:
```
ryan112358, line 5: Remove the try/except, you can assume the user has their environment set up correctly

ryan112358 (commented):
This is a good example but maybe a bit too incomplete for inclusion in the examples/ directory.  I think it would fit better in the module-level documentation of jax_privacy.batch_selection

Issue #88, opened 2026-01-06
Add examples implementing different mechanisms from the literature


ryan112358:
1. [Fine-Tuning Large Language Models with User-Level Differential Privacy](https://arxiv.org/abs/2407.07737)
2. [It's My Data Too: Private ML for Datasets with Multi-User Training Examples](https://arxiv.org/abs/2503.03622)
3. [On Design Principles for Private Adaptive Optimizers](https://arxiv.org/abs/2507.01129)
4. [Differentially Private Adaptive Optimization with Delayed Preconditioners
](https://arxiv.org/abs/2212.00309)
...

amyssnippet:
to crunch on these numbers until I cry binary tears

debanganghosh08:
Hi @ryan112358 ,
I’ve just raised a PR for this issue, implementing four distinct mechanisms from the suggested literature: **ULS** (Charles et al. 2024), **Multi-User Attribution** via _Greedy Bounding_, **Scale-then-Privatize** (Ganesh et al. 2025), and **DP²** (Li et al. 2023).

Regarding the DP² implementation, I’ve structured the training loop to utilize a 'stale' _preconditioner_ that updates every 's' steps using **averaged private gradients**. 
For the Scale-then-Privatize example, I leveraged the _pre_clipping_transform_ hook to **rescale gradients** into a **spherical geometry** before applying the _privacy cli_p.

Could you let me know if this architectural approach is specifically the manual alternating-phase protocol for DP² - does it aligns with your _expectations_ for the library's example suite? 
I’m particularly interested to know if you'd recommend further optimizing the **intra-user averaging** in the _ULS example_ using a more aggressive **vmap** strategy, or if the current **loop-based clarity** is preferred for a _pedagogical_ example. 

I’m looking forward to refine these further based on your standards. 
Thanks!

PR #107, opened 2026-01-20
User-Level Sampling (ULS) for Transformers (#88) Split PR 1

debanganghosh08:
This PR adds a standalone example implementing User-Level Sampling (ULS) for Transformers, as described in Charles et al. (2024), "Fine-Tuning Large Language Models with User-Level Differential Privacy". [Reference](https://arxiv.org/abs/2407.07737) 

**Correctness Verification**: To verify the implementation of the ULS mechanism, I implemented manual logging to ensure that intra-user averaging occurs strictly before the user-level sensitivity clip.

**Averaging Logic**: The script logs the number of sequences per user (e.g., User A: 4 sequences) being collapsed into a single gradient.

**Sensitivity Bounding**: I've added logs for the avg_grad_norm (pre-clip) to verify that the user_clip_norm is correctly applied to the aggregated user update rather than individual records.

This implementation uses a _manual loop_ to handle the varying amounts of _data per user_, ensuring stability in JAX without the need for c_omplex masking of variable-length batches_.

### DIFF
--- a/jax_privacy/experimental/execution_plan.py
+++ b/jax_privacy/experimental/execution_plan.py
@@ -177,11 +177,13 @@ class BandMFExecutionPlanConfig:
 
   Standard DP-SGD is the special case `num_bands=1`. A typical configuration
   for Poisson-sampled DP-SGD is:
+
     - num_bands=1
     - sampling_prob = expected_batch_size / num_examples
     - partition_type = INDEPENDENT
     - neighboring_relation = ADD_OR_REMOVE_ONE
     - accountant = PLDAccountant(ADD_OR_REMOVE_ONE)
+
   This yields the usual Poisson sampling analysis, and uses uncorrelated
   Gaussian noise (the BandMF strategy degenerates to the identity matrix).
 


examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,311 @@
...
+with User-Level Differential Privacy" (https://arxiv.org/abs/2407.07737)
+
+--- Implementation Notes ---
+Standard DP-SGD clips every example. ULS (User-Level Sampling) is different:
+it averages all gradients from a single user into a 'user-update' BEFORE clipping.
+As noted in Sec 4.2 of the paper, this is much more efficient for LLMs because
+it improves the signal-to-noise ratio by reducing the magnitude of the update
+before we hit it with the sensitivity clip.
+
+We use a manual for-loop for the user-aggregation because users often have
```
ryan112358: Please implement this in terms of the following components:

1. jax_privacy.batch_selection.UserSelectionStrategy
2. jax_privacy.clipped_grad(keep_batch_dim=False)

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,311 @@
...
+padding and masking issues that can arise with JAX's vmap.
+"""
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+from jax import random, tree_util
+from jax.flatten_util import ravel_pytree
+import optax
+import tensorflow as tf
```
ryan112358: Can we eliminate the tensorflow dependency?

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # Final projection
+        logits = x @ params["output_projection"]
+        return logits
+
+
+def loss_fn(params, batch):
+    """Cross-entropy loss for language modeling."""
+    inputs, targets = batch[:-1], batch[1:]
+    logits = model.apply_fn(params, inputs)
```
ryan112358: Where does model come from?

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,311 @@
...
+    log_probs = jax.nn.log_softmax(logits)
+    return -jnp.mean(jnp.take_along_axis(log_probs, targets[:, None], axis=-1))
+
+
+def user_level_grad_fn(params, user_data, user_clip_norm):
```
ryan112358: Use jax_privacy.clipped_grad(keep_batch_dim=False) here

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,311 @@
...
+    updates, opt_state = optimizer.update(noisy_grads, opt_state, params)
+    params = optax.apply_updates(params, updates)
+
+    return params, opt_state, noise_state
+
+
+# --- Configuration ---
+config = {
+    "num_steps": 10,
+    "batch_size": 2,  # Number of users per batch
```
ryan112358: Let's call batch_size --> users_per_batch and add a new config option for examples_per_user

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,311 @@
...
+)
+
+opt_state = optimizer.init(params)
+noise_state = privatizer.init(params)
+
+# --- Training Loop ---
+print("Starting user-level DP training...")
+for step in range(config["num_steps"]):
+    # Sample a batch of users for this step
+    sampled_user_ids = np.random.choice(
```
ryan112358: Let's use jax_privacy.batch_selection, sampling users with poisson sampling and selecting examples by using the UserSelectionStrategy

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,311 @@
...
+        self.id_to_char = {i: c for i, c in enumerate(self.vocab)}
+
+    def encode(self, text):
+        return np.array([self.char_to_id[c] for c in text], dtype=np.int32)
+
+    def decode(self, ids):
+        return "".join([self.id_to_char[i] for i in ids])
+
+
+class TransformerDecoder:
```
ryan112358: As we begin to add more example scripts, there is some redundancy in the model definition and data loading.  I wonder if you can have a single standard model definition (maybe written in flax nnx or equinox), and then just import it here and in the other scripts?

examples/dp_sgd_transformer_nnx.py
```diff
@@ -0,0 +1,369 @@
...
+  # and attention expectations which look for a 4D mask derived from
+  # a batched input. We index [0] on the output to remove it.
+  # x indices: [seq_len] -> [1, seq_len]
+  logits = m(x[None, :])[0]
+
+
+  return optax.softmax_cross_entropy_with_integer_labels(logits, y).mean()
+
+
+def main():
```
Neerajpathak07:
```suggestion
def main(argv: Sequence[str]) -> None:
  if len(argv) > 1:
    raise app.UsageError('Too many command-line arguments.')
```
feel free to add :-
```python
from typing import Sequence
```
if you think this suggestion makes sense!!!

examples/dp_sgd_transformer_nnx.py
```diff
@@ -0,0 +1,369 @@
...
+  # Configure DP Gradient Clipping
+  grad_fn = clipped_grad(
+      functools.partial(pure_loss_fn, graphdef=graphdef, other=other),
+      l2_clip_norm=CLIP_NORM,
+      batch_argnums=(1, 2),  # x and y are batched
+      keep_batch_dim=False,  # Process per-example
+      return_values=True     # Return loss values for logging
+  )
+
+  privatizer = noise_addition.gaussian_privatizer(
```
Neerajpathak07: I believe you'll have to update this as `noise_addition` should be wired through `plan` which in turn has to be configured first.
Ref:- https://github.com/google-deepmind/jax_privacy/blob/84daf71e6e7070edd3e1e16245723e8a60dcd062/examples/dp_logistic_regression.py#L88

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,182 @@
...
+    'users_per_batch', 4, 'Number of users to select in each batch.'
+)
+_EXAMPLES_PER_USER = flags.DEFINE_integer(
+    'examples_per_user', 2, 'Number of examples to select for each user.'
+)
+_STEPS = flags.DEFINE_integer('steps', 10, 'Number of training steps.')
+_L2_CLIP_NORM = flags.DEFINE_float(
+    'l2_clip_norm', 1.0, 'L2 clipping norm for gradients.'
+)
+_LEARNING_RATE = flags.DEFINE_float('learning_rate', 1e-3, 'Learning rate.')
```
Neerajpathak07: Hey, @debanganghosh08 instead of flagging it like this here let's create hyperparameters for this and provide the apt value.
something similar has also been done in the upstream versions.
Ref:- https://github.com/google-deepmind/jax_privacy/blob/84daf71e6e7070edd3e1e16245723e8a60dcd062/examples/dp_logistic_regression.py#L28

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,182 @@
...
+  num_users = 20
+  num_examples_per_user = np.random.randint(1, 10, size=num_users)
+
+  data, labels, user_ids = get_synthetic_data(
+      num_users=num_users,
+      num_examples_per_user=num_examples_per_user,
+      seq_len=seq_len,
+      vocab_size=vocab_size,
+  )
+
```
Neerajpathak07, line 114: @ryan112358 Correct me if I am wrong but one thing that I have learned is that while applying poisson sampling for any example the best practice is to jot it down from `execution_plan.BandMFExecutionPlanConfig` and create a `plan` var to clipp the grad function.
Maybe we can do something similar here?

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,182 @@
...
+  model = TransformerDecoder(
+      vocab_size=vocab_size,
+      embed_dim=embed_dim,
+      num_heads=num_heads,
+      ff_dim=ff_dim,
+  )
+  params = model.init(
+      jax.random.key(0), jnp.zeros((1, seq_len), dtype=jnp.int32), train=False
+  )['params']
+  optimizer = optax.adam(_LEARNING_RATE.value)
```
Neerajpathak07:
```suggestion
  optimizer = optax.sgd(LEARNING_RATE)
```
where `LEARNING_RATE` will be the hyperparameter value hardcoded at top.

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,182 @@
...
+  opt_state = optimizer.init(params)
+
+  # 2. Batch Selection
+  sampling_prob = _USERS_PER_BATCH.value / num_users
```
Neerajpathak07:
```suggestion
  sampling_prob = EXPECTED_BATCH_SIZE / USERS
```
maybe we can update this for `EXPECTED_BATCH_SIZE` and add it as an argument for `config`
ref:- https://github.com/google-deepmind/jax_privacy/blob/84daf71e6e7070edd3e1e16245723e8a60dcd062/examples/dp_logistic_regression.py#L75

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,182 @@
...
+  @jax.jit
+  def train_step(params, opt_state, batch_data, batch_labels):
+    grads = grad_fn(params, batch_data, batch_labels)
+    updates, opt_state = optimizer.update(grads, opt_state, params)
+    params = optax.apply_updates(params, updates)
+    return params, opt_state
+
+  # 4. Training Loop
+  batch_iterator = user_strategy.batch_iterator(user_ids, rng=0)
+  for step, user_batch_indices in enumerate(batch_iterator):
```
Neerajpathak07, line 187:
```suggestion
  for step, user_batch_indices in enumerate(plan.batch_selection_strategy.batch_iterator(USERS)):
```
for poisson sampling this should come from `batch_selection_strategy`

examples/dp_sgd_transformer_nnx.py
debanganghosh08: Thanks for the suggestion, Neeraj! Adding the Sequence[str] type hint and the UsageError check makes the entry point much more robust. I'll include this in the next push.

examples/user_level_transformer_example.py
debanganghosh08: Good point on the configuration style. Moving from flags to hardcoded constants at the top aligns better with the dp_logistic_regression template. I'll refactor the config block to use constants like USERS and LEARNING_RATE.

examples/user_level_transformer_example.py
debanganghosh08: I used adam here because it's the standard for Transformer convergence, but I'm happy to switch to sgd if we want to keep the example as simple as possible. @ryan112358, do you have a preference for the optimizer type in this example?

examples/user_level_transformer_example.py
debanganghosh08: Regarding the execution_plan and BandMFExecutionPlanConfig: I see how this is used in the upstream logistic regression example. However, for this Transformer example, I implemented the loop manually to make the interaction between UserSelectionStrategy and clipped_grad explicit for users. Do you think moving to the plan abstraction provides enough benefit to outweigh the loss of visibility into the DP mechanics for this specific use case?

examples/user_level_transformer_example.py
Neerajpathak07: here the best thing to do is compare what both of them do.

`Adam`-> Adam is an SGD variant with gradient scaling adaptation. The scaling used for each parameter is computed from estimates of first and second-order moments of the gradients (using suitable exponential moving averages).

`sgd` -> A canonical Stochastic Gradient Descent optimizer.
This implements stochastic gradient descent. It also includes support for momentum, and Nesterov acceleration, as these are standard practice when using stochastic gradient descent to train deep neural networks.

this should give you a broad idea of what would be a better fit here.

examples/user_level_transformer_example.py
Neerajpathak07: Yes, apparently we are not just clipping the grads just for users, we also want this for `sampling_prob` and other such parameters. Having them consisted at `clipped_grad` and wiring poisson sampling is the Jax_privacy way of protecting the dp mechanics.

examples/user_level_transformer_example.py
ryan112358: The ExecutionPlanConfig can be used or ignored, up to you.  It's kind of in an experimental state right now and the API might change in the future

examples/user_level_transformer_example.py
ryan112358: +1 to keeping adam here (or maybe switching to adamw).  Although might be worth trying out both as Neeraj suggests and seeing which one works better

examples/dp_sgd_transformer_nnx.py
```diff
@@ -0,0 +1,390 @@
...
+  # graph and other state.
+  graphdef, params, other = nnx.split(model, nnx.Param, ...)
+  opt_state = optimizer.init(params)
+
+  # Configure DP Gradient Clipping
+  grad_fn = clipped_grad(
+      functools.partial(pure_loss_fn, graphdef=graphdef, other=other),
+      l2_clip_norm=CLIP_NORM,
+      batch_argnums=(1, 2),  # x and y are batched
+      keep_batch_dim=True,   # Return per-example gradients
```
ryan112358, line 312: Delete this inline comment, it's not quite right.  All it means is when computing per-example gradients, pass in batches with a leading axis size of (1,), rather than stripping it out completely.  

Actually you can just delete the whole line since this is the default value

examples/dp_sgd_transformer_nnx.py
```diff
@@ -0,0 +1,390 @@
...
+    return params, opt_state, noise_state, loss.values.mean()
+
+  # Training loop
+  print(f'Training for {NUM_STEPS} steps...')
+  iterator = plan.batch_selection_strategy.batch_iterator(dataset_size)
+  for step, batch_indices in enumerate(iterator):
+    if step >= NUM_STEPS:
+      break
+
+    if len(batch_indices) == 0:
```
ryan112358, line 372: Technically you still need to add noise even when you have an empty batch.  If removing this lin breaks something internal to Jax privacy downstream, feel free to leave a TODO for now and we can fix it in follow up.  It it doesn't break anything, then just delete this if -> continue block

examples/dp_sgd_transformer_nnx.py
```diff
@@ -0,0 +1,390 @@
...
+      continue
+
+    # Construct batch from indices
+    x = np.stack([data[i : i + CONTEXT_LENGTH] for i in batch_indices])
```
ryan112358, line 376: This logic doesn't look right to me.  For sequence models, data should have shape (N, CONTEXT_LENGTH), and you should be able to do something like data[batch_indices] to get something with shape (B, CONTEXT_LENGTH)

examples/dp_sgd_transformer_nnx.py
```diff
@@ -0,0 +1,390 @@
...
+      num_bands=1,
+      epsilon=EPSILON,
+      delta=DELTA,
+      sampling_prob=BATCH_SIZE / dataset_size,
+  )
+  plan = config.make(grad_fn)
+  privatizer = plan.noise_addition_transform
+  noise_state = privatizer.init(params)
+
+  @jax.jit
```
ryan112358, line 331: For efficiency you can donate some argnums (params, opt_state, noise_state)

examples/dp_sgd_transformer_nnx.py
```diff
@@ -0,0 +1,390 @@
...
+    text: The text to encode.
+    stoi: The string-to-index mapping.
+
+  Returns:
+    A numpy array of integer indices.
+  """
+  return np.array([stoi[ch] for ch in text], dtype=np.int32)
+
+
+def get_batch(
```
ryan112358, line 102: This function appears unused and can be deleted.

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,209 @@
...
+    x = nn.Dense(self.vocab_size)(x)
+    return x
+
+
+def get_synthetic_data(
+    num_users: int,
+    num_examples_per_user: list[int],
+    seq_len: int,
+    vocab_size: int,
+):
```
ryan112358, line 78: Can you document the shapes of the output arrays?

examples/user_level_transformer_example.py
```diff
@@ -0,0 +1,209 @@
...
+
+  # Wrap the plan's strategy with UserSelectionStrategy
+  # We assume plan.batch_selection_strategy is compatible
+  # (CyclicPoissonSampling)
+  user_strategy = UserSelectionStrategy(
+      base_strategy=plan.batch_selection_strategy,
+      examples_per_user_per_batch=EXAMPLES_PER_USER,
+  )
+
+  @jax.jit
```
ryan112358, line 171: You could donate_argnums here as well

jax_privacy/experimental/execution_plan.py
```diff
@@ -177,11 +177,13 @@ class BandMFExecutionPlanConfig:
 
   Standard DP-SGD is the special case `num_bands=1`. A typical configuration
   for Poisson-sampled DP-SGD is:
+
```
ryan112358, line 180: I think this file changed, please rebase to head and revert this change

PR #196, opened 2026-03-20
Make Poisson sampling opt-in in the Keras DP path

staticpayload:
## What this PR changes

This PR updates the Keras DP path so Poisson sampling can be done directly inside `fit()` without breaking the existing API.

The main change is an opt-in flag, `DPKerasConfig.poisson_sampling_in_fit`, which defaults to `False`. When it is enabled, the wrapper builds a Poisson-sampled `keras.utils.PyDataset` from random-access per-example arrays, validates the input sizes against the DP config, and rejects `validation_split` because the accountant needs the exact training-set size.

## Main changes

- Add `poisson_sampling_in_fit` to `DPKerasConfig`, defaulting to `False` for backward compatibility.
- Build Poisson-sampled batches internally only when that flag is enabled.
- Validate random-access array inputs and ensure `x`, `y`, and `sample_weight` agree on batch size in the Poisson-in-fit path.
- Represent padded examples through zero sample weights and derive the padding mask from `sample_weight` inside the DP training path.
- Update the Keras example and docs to show the opt-in Poisson path explicitly.
- Expand Keras tests around batch construction, padding, validation errors, and fit-argument rewriting.
- Add the empty-input padding edge case coverage in `batch_selection`.

## Why this shape

The goal here is to make the privacy-sensitive Poisson-sampling path available directly in the Keras wrapper while keeping the older workflows intact. Users who already batch or sample outside the wrapper can keep doing that. Users who want the wrapper to own batch formation can enable the flag and get a path whose sampling behavior matches the DP accounting assumptions.

## Verification

I ran the following locally in Python 3.11:

- `KERAS_BACKEND=jax python -m pytest tests/keras_api_test.py -q`
- `python -m pytest tests/batch_selection_test.py -q -k 'pad_to_multiple_of'`
- `python -m pyink --check --diff jax_privacy/keras_api.py tests/keras_api_test.py`
- `python -m pylint --rcfile=.pylintrc jax_privacy/keras_api.py`

### DIFF
--- a/jax_privacy/keras_api.py
+++ b/jax_privacy/keras_api.py
@@ -45,6 +45,7 @@ from collections.abc import Callable
 import dataclasses
 import functools
 import inspect
+import math
 import types
 from typing import Any
 
@@ -52,6 +53,7 @@ import chex
 import jax
 import jax.numpy as jnp
 import jax_privacy
+from jax_privacy import batch_selection
 from jax_privacy.accounting import accountants
 from jax_privacy.accounting import analysis
 from jax_privacy.accounting import calibrate
@@ -78,7 +80,10 @@ class DPKerasConfig:
         privacy guarantees you have to achieve. You should not increase the
         delta only because of poor model performance.
       clipping_norm: The clipping norm for the per-example gradients.
-      batch_size: The batch size for the training.
+      batch_size: The batch size used by the DP optimizer. When
+        `poisson_sampling_in_fit=True`, this is the expected batch size of the
+        internal Poisson sampler. Otherwise it must match the batch size
+        supplied to `fit()`.
       gradient_accumulation_steps: The number of gradient accumulation steps.
         This is the number of batches to accumulate before adding noise and
         performing an optimizer step. 1 means that there is no gradient
@@ -99,6 +104,10 @@ class DPKerasConfig:
       train_size: The number of training examples in the dataset. If you repeat
         the examples in your dataset iterator, it should be the number of
         training examples in the original dataset before repeating.
+      poisson_sampling_in_fit: Whether `fit()` should internally resample
+        random-access array inputs using Poisson sampling. Leave this as False
+        for backwards-compatible behavior or when the user supplies a dataset
+        iterator that already handles sampling.
       noise_multiplier: The noise multiplier for the gradients. If None
         (recommended), the noise multiplier will be automatically calculated
         based on epsilon, delta, effective_batch_size, train_steps and
@@ -133,6 +142,7 @@ class DPKerasConfig:
   gradient_accumulation_steps: int
   train_steps: int
   train_size: int
+  poisson_sampling_in_fit: bool = False
   noise_multiplier: float | None = None
   rescale_to_unit_norm: bool = True
   microbatch_size: int | None = None
@@ -194,15 +204,30 @@ class DPKerasConfig:
       raise ValueError(f'Clipping norm {self.clipping_norm} must be positive.')
     if self.batch_size <= 0:
       raise ValueError(f'Batch size {self.batch_size} must be positive.')
-    if self.train_steps <= 0:
-      raise ValueError(f'Train steps {self.train_steps} must be positive.')
     if self.train_size <= 0:
       raise ValueError(f'Train size {self.train_size} must be positive.')
+    if self.batch_size > self.train_size:
+      raise ValueError(
+          f'Batch size {self.batch_size} must be less than or equal to train'
+          f' size {self.train_size}.'
+      )
+    if self.train_steps <= 0:
+      raise ValueError(f'Train steps {self.train_steps} must be positive.')
     if self.gradient_accumulation_steps <= 0:
       raise ValueError(
           f'Gradient accumulation steps {self.gradient_accumulation_steps} must'
           ' be positive.'
       )
+    if self.microbatch_size is not None:
+      if self.microbatch_size <= 0:
+        raise ValueError(
+            f'Microbatch size {self.microbatch_size} must be positive.'
+        )
+      if self.microbatch_size > self.batch_size:
+        raise ValueError(
+            f'Microbatch size {self.microbatch_size} must be less than or'
+            f' equal to batch size {self.batch_size}.'
+        )
     if self.noise_multiplier is not None:
       if self.noise_multiplier <= 0:
         raise ValueError(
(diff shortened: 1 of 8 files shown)


ryan112358 (approved):

jax_privacy/keras_api.py
```diff
@@ -77,8 +79,11 @@ class DPKerasConfig:
         noise). You should set this value before training and only based on the
         privacy guarantees you have to achieve. You should not increase the
         delta only because of poor model performance.
-      clipping_norm: The clipping norm for the per-example gradients.
-      batch_size: The batch size for the training.
+      clipping_norm: The clipping norm for the gradients.
```
dvadym: whe per-example was removed? clipping is performed per each example

jax_privacy/keras_api.py
```diff
@@ -353,6 +380,218 @@ def _add_dp_sgd_attributes(model: keras.Model, params: DPKerasConfig) -> None:
...
+  leaves = jax.tree.leaves(tree)
+  if not leaves:
+    raise ValueError('Expected at least one array leaf in the training data.')
+  batch_size = None
+  for leaf in leaves:
+    if not hasattr(leaf, 'shape'):
+      raise ValueError(
+          'DP Keras training requires random-access array-like inputs.'
+      )
+    if len(leaf.shape) == 0:
```
dvadym: Nit (style improvement): 

```
if not leaf.shape: 

```
https://google.github.io/styleguide/pyguide.html#2144-decision

jax_privacy/keras_api.py
```diff
@@ -353,6 +380,218 @@ def _add_dp_sgd_attributes(model: keras.Model, params: DPKerasConfig) -> None:
...
+    elif leaf_batch_size != batch_size:
+      raise ValueError(
+          'All training data leaves must agree on their leading batch'
+          ' dimension.'
+      )
+  return int(batch_size)
+
+
+def _take_batch_from_leaf(leaf: chex.Array, indices: np.ndarray) -> np.ndarray:
+  """Slices one array leaf, turning padded ``-1`` indices into zeros."""
```
dvadym: could you please clarify the docstring

jax_privacy/keras_api.py
```diff
@@ -353,6 +380,218 @@ def _add_dp_sgd_attributes(model: keras.Model, params: DPKerasConfig) -> None:
...
+  if sample_weight.ndim == 2:
+    return ~jnp.any(sample_weight, axis=1)
+  raise ValueError('Expected sample_weight to be a 1D or 2D array.')
+
+
+def _maybe_symbolically_build_private_model(
+    model: keras.Model, dataset: _PoissonSampledTrainingDataset
+) -> None:
+  """Runs Keras' symbolic build before fit() sees the Poisson dataset."""
+  if not hasattr(model, '_symbolic_build'):
```
dvadym, line 575: what if we don't symbolically build?

jax_privacy/keras_api.py
```diff
@@ -353,6 +380,218 @@ def _add_dp_sgd_attributes(model: keras.Model, params: DPKerasConfig) -> None:
...
+def _pad_batch_indices(indices: np.ndarray, multiple: int) -> np.ndarray:
+  """Pads indices with -1 so empty Poisson draws are still representable."""
+  if indices.size == 0:
+    return np.full(multiple, -1, dtype=np.int32)
+  return batch_selection.pad_to_multiple_of(indices, multiple)
+
+
+def _tree_batch_size(tree: chex.ArrayTree) -> int:
+  """Returns and validates the batch size of a pytree of arrays."""
+  leaves = jax.tree.leaves(tree)
```
dvadym, line 498: Please write down in comment expectations on `tree` (as far as I understood: all leaves have the same 0-th dimenstions

jax_privacy/keras_api.py
```diff
@@ -671,11 +989,16 @@ def _noised_clipped_grads(
 
   noisy_grads, new_noise_state = privatizer.update(clipped_grad, noise_state)
 
-  loss = per_example_aux.values.mean()
-  unscaled_loss = per_example_aux.aux[0].mean()
+  # Open question: should these aggregates be means or sums?
```
dvadym: mean

jax_privacy/keras_api.py
```diff
@@ -353,6 +380,218 @@ def _add_dp_sgd_attributes(model: keras.Model, params: DPKerasConfig) -> None:
...
+
+
+def _masked_mean(
+    values: chex.Array, is_padding_example: jax.Array
+) -> chex.Array:
+  """Averages only the non-padding examples, returning 0 for empty batches."""
+  values = jnp.asarray(values)
+  if values.ndim == 0:
+    return values
+  where = jnp.asarray(~is_padding_example)
```
dvadym, line 594: Nit (style improvement)
```
retturn values[~is_padding_example, :].mean(axis=0)
```

dvadym (commented):
Thanks it looks cool. I've left some comments.

jax_privacy/keras_api.py
chai1any: Good catch. I restored “per-example gradients” there. The clipping is still per example; I had simplified the wording too aggressively. My bad

jax_privacy/keras_api.py
chai1any: Applied. I switched that check to if not leaf.shape.

jax_privacy/keras_api.py
chai1any: It now says that the helper slices a batched leaf and zero-fills padded -1 index positions.

jax_privacy/keras_api.py
chai1any: I expanded the comment there. The reason for the eager symbolic build is that otherwise a lazily-built model can hit the wrapped fit path before Keras has created the model and optimizer state from a standard batch.

jax_privacy/keras_api.py
chai1any: Added. There is now an inline comment saying the helper expects a non-empty pytree of random-access arrays whose leaves all share the same leading batch dimension.

jax_privacy/keras_api.py
chai1any: Agreed. I removed the open-question wording and made the masked-mean reduction explicit in the comments.

jax_privacy/keras_api.py
chai1any: I left this one as-is on purpose. The boolean-indexing version is simpler to read, but it introduces a data-dependent shape, which is risky under JAX JIT and sharding. The current jnp.mean(..., where=...) path keeps the shape static while doing the same masked reduction.

dvadym: Goot point.

dvadym (approved):
Thanks!

staticpayload:
To make this concrete: this was a force-push, not an additive commit on top of the old branch.

GitHub recorded the PR head moving from:
- `c74f7956ed1a2307117ccabc535845297e096b1c`

to:
- `944a6144bbf29fcfd394eb945dd35367f2396260`

at `2026-04-01T13:35:58Z`.

So the branch did move; it just moved via a history rewrite onto the latest `main`, which is why there is now a single replacement commit instead of an extra commit appended on top.

ryan112358 (approved):

ryan112358:
@staticpayload It seems there was a regression in keras_api_example, can you look into it?

FYI @RamSaw 

  Traceback (most recent call last):
  DP: expected train accuracy: ~96%, val accuracy: ~92%
    File "/home/runner/work/jax_privacy/jax_privacy/examples/keras_api_example.py", line 138, in <module>
  Non-DP: expected train accuracy: ~98%, val accuracy: ~98%
      app.run(main)
    File "/opt/hostedtoolcache/Python/3.11.15/x64/lib/python3.11/site-packages/absl/app.py", line 367, in run
      _run_main(main, args)
    File "/opt/hostedtoolcache/Python/3.11.15/x64/lib/python3.11/site-packages/absl/app.py", line 312, in _run_main
      sys.exit(main(argv))
               ^^^^^^^^^^
    File "/home/runner/work/jax_privacy/jax_privacy/examples/keras_api_example.py", line 129, in main
      final_accuracy > 0.85
  AssertionError: DP Accuracy 0.8287 is too low!

PR #218, opened 2026-04-11
Add banded-inverse methods

### LINKED ISSUE #202: Add support for BandInvMF / BISR factorizations from "Back to Square Roots" ICLR 2026 paper
The recent paper [Back to square roots: An optimal bound on the matrix factorization error for multi-epoch differentially private SGD](https://arxiv.org/pdf/2505.12128) introduces banded inverse factorizations **BandInvMF** and **BISR**, where the correlation matrix $C^{-1}$ is banded rather than the strategy matrix $C$.
 It would be nice to add support for these here.

Plan:

Add a new `banded_inverse.py` file with the following functions:

- `optimize_banded_inverse_toeplitz`
- `compute_banded_inverse_square_root`
- `compute_banded_inverse_sensitivity`

`optimize_banded_inverse_toeplitz` should optimize over banded inverse Toeplitz matrices. This will likely reuse the following functionality from `toeplitz.py`:

- `minsep_sensitivity_squared`
- `inverse_coef`
- `per_query_error`

and `optimize` from `optimization.py`.

For `compute_banded_inverse_sensitivity`, note that **BandInvMF** is not necessarily decreasing. A safe option is to upper bound the sensitivity using the corresponding decreasing sequence suggested in the paper, which then only requires `toeplitz.minsep_sensitivity_squared`. A tighter alternative would be to use `sensitivity.get_min_sep_sensitivity_upper_bound_for_X`, which could be enabled with a flag.

`compute_banded_inverse_square_root` should implement the explicit **BISR** construction, which can be built efficiently without using the rest of the library.

npkalinin:
Resolves #202

This PR adds support for BandInvMF and BISR, based on banded inverse lower-triangular Toeplitz matrices, from https://arxiv.org/pdf/2505.12128.

Changes:
- add `banded_inverse_square_root_noising_coefs(...)` in `toeplitz.py`
- add `compute_banded_inverse_sensitivity(...)` for banded inverse lower-triangular Toeplitz noising matrices
- add `optimize_banded_inverse_toeplitz(...)` for direct optimization over banded inverse lower-triangular Toeplitz noising coefficients
- extend `examples/dpmf_strategy_optimization.py` with:
  - `banded-inverse-toeplitz`
  - `banded-inverse-sqrt`

### DIFF
--- a/examples/dpmf_strategy_optimization.py
+++ b/examples/dpmf_strategy_optimization.py
@@ -108,20 +108,11 @@ def optimize_strategy(
 
     case 'banded-inverse-toeplitz':
       # https://arxiv.org/pdf/2505.12128
-      weight_decay = 1.0
-      momentum = 0.0
-      workload_coef = toeplitz.multiply(
-          weight_decay ** jnp.arange(n),
-          momentum ** jnp.arange(n),
-          n=n,
-          skip_checks=True,
-      )
       noising_coef = toeplitz.optimize_banded_inverse_toeplitz(
           n=n,
           min_sep=sep,
           num_bands=sep,
-          weight_decay=weight_decay,
-          momentum=momentum,
+          max_participations=participations,
           max_optimizer_steps=1000,
           reduction_fn=reduction_fn,
       )
@@ -132,7 +123,7 @@ def optimize_strategy(
           max_participations=participations,
       )
       pqe = toeplitz.per_query_error(
-          noising_coef=noising_coef, n=n, workload_coef=workload_coef
+          noising_coef=noising_coef, n=n, workload_coef=jnp.ones(n)
       )
       loss = reduction_fn(pqe) * sensitivity_squared
 
@@ -172,17 +163,7 @@ def optimize_strategy(
 
     case 'banded-inverse-sqrt':
       # https://arxiv.org/pdf/2505.12128
-      weight_decay = 1.0
-      momentum = 0.0
-      workload_coef = toeplitz.multiply(
-          weight_decay ** jnp.arange(n),
-          momentum ** jnp.arange(n),
-          n=n,
-          skip_checks=True,
-      )
-      noising_coef = toeplitz.banded_inverse_square_root_noising_coefs(
-          sep, weight_decay=weight_decay, momentum=momentum
-      )
+      noising_coef = toeplitz.banded_inverse_square_root_noising_coefs(sep)
       sensitivity_squared = toeplitz.compute_banded_inverse_sensitivity_squared(
           n=n,
           noising_coef=noising_coef,
@@ -190,7 +171,7 @@ def optimize_strategy(
           max_participations=participations,
       )
       pqe = toeplitz.per_query_error(
-          noising_coef=noising_coef, n=n, workload_coef=workload_coef
+          noising_coef=noising_coef, n=n, workload_coef=jnp.ones(n)
       )
       loss = reduction_fn(pqe) * sensitivity_squared
 
--- a/examples/dpmf_strategy_optimization.py
+++ b/examples/dpmf_strategy_optimization.py
@@ -130,7 +130,7 @@ def optimize_strategy(
           noising_coef=noising_coef,
           min_sep=sep,
           max_participations=participations,
-      ) 
+      )
       pqe = toeplitz.per_query_error(
           noising_coef=noising_coef, n=n, workload_coef=workload_coef
       )
@@ -164,7 +164,6 @@ def optimize_strategy(
       # https://arxiv.org/abs/2202.11205
       # https://arxiv.org/abs/2405.13763
       strategy_coef = toeplitz.optimal_max_error_strategy_coefs(sep)
-      strategy_coef = jnp.concatenate([strategy_coef, jnp.zeros(n - sep)])
       sensitivity_squared = toeplitz.minsep_sensitivity_squared(
           strategy_coef, min_sep=sep, max_participations=participations
       )
--- a/examples/dpmf_strategy_optimization.py
+++ b/examples/dpmf_strategy_optimization.py
@@ -50,8 +50,10 @@ _STRATEGY = flags.DEFINE_enum(
         'dense',
         'banded',
         'banded-toeplitz',
+        'banded-inverse-toeplitz',
         'blt',
         'banded-sqrt',
+        'banded-inverse-sqrt',
         'normalized-banded-toeplitz',
     ],
     'Strategy class to optimize over.  See code for details.',
@@ -104,6 +106,35 @@ def optimize_strategy(
       pqe = toeplitz.per_query_error(strategy_coef=strategy_coef, n=n)
       loss = reduction_fn(pqe) * sensitivity_squared
 
+    case 'banded-inverse-toeplitz':
+      # https://arxiv.org/pdf/2505.12128
+      weight_decay = 1.0
+      momentum = 0.0
+      workload_coef = toeplitz.multiply(
+          weight_decay**jnp.arange(n),
+          momentum**jnp.arange(n),
+          n=n,
+          skip_checks=True,
+      )
+      noising_coef = toeplitz.optimize_banded_inverse_toeplitz(
+          n=n,
+          min_sep=sep,
+          num_bands=sep,
+          weight_decay=weight_decay,
+          momentum=momentum,
+          max_optimizer_steps=10,
+      )
+      sensitivity_squared = toeplitz.compute_banded_inverse_sensitivity(
+          n=n,
+          noising_coef=noising_coef,
+          min_sep=sep,
+          max_participations=participations,
+      ) ** 2
+      pqe = toeplitz.per_query_error(
+          noising_coef=noising_coef, n=n, workload_coef=workload_coef
+      )
+      loss = reduction_fn(pqe) * sensitivity_squared
+
     case 'normalized-banded-toeplitz':
       # https://arxiv.org/abs/2405.15913
       def loss_fn(coef):  # pylint: disable=function-redefined
@@ -136,6 +167,30 @@ def optimize_strategy(
           strategy_coef, min_sep=sep, max_participations=participations
       )
       pqe = toeplitz.per_query_error(strategy_coef=strategy_coef, n=n)
+      loss = reduction_fn(pqe) * sensitivity_squared
+
+    case 'banded-inverse-sqrt':
+      # https://arxiv.org/pdf/2505.12128
+      weight_decay = 1.0
+      momentum = 0.0
+      workload_coef = toeplitz.multiply(
+          weight_decay**jnp.arange(n),
+          momentum**jnp.arange(n),
+          n=n,
+          skip_checks=True,
+      )
+      noising_coef = toeplitz.banded_inverse_square_root_noising_coefs(
+          sep, weight_decay=weight_decay, momentum=momentum
+      )
+      sensitivity_squared = toeplitz.compute_banded_inverse_sensitivity(
+          n=n,
+          noising_coef=noising_coef,
+          min_sep=sep,
+          max_participations=participations,
+      ) ** 2
+      pqe = toeplitz.per_query_error(
+          noising_coef=noising_coef, n=n, workload_coef=workload_coef
+      )
       loss = reduction_fn(pqe) * sensitivity_squared
 
     case 'banded':
--- a/examples/dpmf_strategy_optimization.py
+++ b/examples/dpmf_strategy_optimization.py
@@ -111,8 +111,8 @@ def optimize_strategy(
       weight_decay = 1.0
       momentum = 0.0
       workload_coef = toeplitz.multiply(
-          weight_decay**jnp.arange(n),
-          momentum**jnp.arange(n),
+          weight_decay ** jnp.arange(n),
+          momentum ** jnp.arange(n),
           n=n,
           skip_checks=True,
       )
@@ -122,14 +122,15 @@ def optimize_strategy(
           num_bands=sep,
           weight_decay=weight_decay,
           momentum=momentum,
-          max_optimizer_steps=10,
+          max_optimizer_steps=1000,
+          reduction_fn=reduction_fn,
       )
-      sensitivity_squared = toeplitz.compute_banded_inverse_sensitivity(
+      sensitivity_squared = toeplitz.compute_banded_inverse_sensitivity_squared(
           n=n,
           noising_coef=noising_coef,
           min_sep=sep,
           max_participations=participations,
-      ) ** 2
+      ) 
       pqe = toeplitz.per_query_error(
           noising_coef=noising_coef, n=n, workload_coef=workload_coef
       )
@@ -163,6 +164,7 @@ def optimize_strategy(
       # https://arxiv.org/abs/2202.11205
       # https://arxiv.org/abs/2405.13763
       strategy_coef = toeplitz.optimal_max_error_strategy_coefs(sep)
+      strategy_coef = jnp.concatenate([strategy_coef, jnp.zeros(n - sep)])
       sensitivity_squared = toeplitz.minsep_sensitivity_squared(
           strategy_coef, min_sep=sep, max_participations=participations
       )
@@ -174,20 +176,20 @@ def optimize_strategy(
       weight_decay = 1.0
       momentum = 0.0
       workload_coef = toeplitz.multiply(
-          weight_decay**jnp.arange(n),
-          momentum**jnp.arange(n),
+          weight_decay ** jnp.arange(n),
+          momentum ** jnp.arange(n),
           n=n,
           skip_checks=True,
       )
       noising_coef = toeplitz.banded_inverse_square_root_noising_coefs(
           sep, weight_decay=weight_decay, momentum=momentum
       )
-      sensitivity_squared = toeplitz.compute_banded_inverse_sensitivity(
+      sensitivity_squared = toeplitz.compute_banded_inverse_sensitivity_squared(
           n=n,
           noising_coef=noising_coef,
           min_sep=sep,
           max_participations=participations,
-      ) ** 2
+      )
       pqe = toeplitz.per_query_error(
           noising_coef=noising_coef, n=n, workload_coef=workload_coef
       )
--- a/jax_privacy/matrix_factorization/toeplitz.py
+++ b/jax_privacy/matrix_factorization/toeplitz.py
@@ -736,6 +736,7 @@ def optimize_coefs_for_amplifications(
   return coef, stddev
 
 
+@functools.partial(jax.jit, static_argnums=[0])
 def banded_inverse_square_root_noising_coefs(
     num_bands: int,
     workload_coef: jax.Array | None = None,
@@ -837,6 +838,7 @@ def optimize_banded_inverse_toeplitz(
     n: int,
     min_sep: int,
     num_bands: int,
+    *,
     noising_coef: jax.Array | None = None,
     strategy_coef: jax.Array | None = None,
     workload_coef: jax.Array | None = None,
--- a/jax_privacy/matrix_factorization/toeplitz.py
+++ b/jax_privacy/matrix_factorization/toeplitz.py
@@ -775,7 +775,8 @@ def compute_banded_inverse_sensitivity_squared(
     max_participations: int | None = None,
     use_matrix_upper_bound: bool = False,
 ) -> jax.Array:
-  """Returns the squared sensitivity of a banded inverse Toeplitz noising matrix.
+  """Returns the squared sensitivity of a banded inverse Toeplitz noising
+  matrix.
 
   This function takes Toeplitz coefficients of a lower-triangular noising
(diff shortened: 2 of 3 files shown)


jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,167 @@ def optimize_coefs_for_amplifications(
...
+            workload_coef=workload_coef,
+            skip_checks=True,
+        )
+    )
+    sens_squared = minsep_sensitivity_squared(
+        strategy_coef,
+        min_sep=min_sep,
+        max_participations=None,
+        n=n,
+        skip_checks=True,
```
chai1any: I think this objective is using the risky sensitivity path.

`minsep_sensitivity_squared(...)` is the closed-form Toeplitz sensitivity expression, and when checks are enabled it explicitly expects non-negative, non-increasing coefficients. Here `strategy_coef = solve_banded(coef, e_0)` is unconstrained, so during optimization it can easily leave that regime while we still optimize against `minsep_sensitivity_squared(..., skip_checks=True)` as if the assumptions still held.

That seems different from the safe BandInvMF route described in the paper and in this PR docstring, where the robust path is to work with `abs(strategy_coef)` and, when needed, upper-bound it by the smallest non-increasing majorant before using the Toeplitz sensitivity formula.

I would suggest factoring out a JAX-native helper for the optimizer, something like:

```python
def _banded_inverse_sensitivity_squared_upper_bound(
    *,
    n: int,
    noising_coef: jax.Array,
    min_sep: int,
    max_participations: int | None = None,
) -> jax.Array:
  strategy_coef = inverse_coef(noising_coef, n)
  coef = jax.lax.cummax(jnp.abs(strategy_coef)[::-1])[::-1]
  return minsep_sensitivity_squared(
      coef,
      min_sep=min_sep,
      max_participations=max_participations,
      n=n,
      skip_checks=True,
  )

File: jax_privacy/matrix_factorization/toeplitz.py [hunk 2]
```diff
@@ -734,3 +734,167 @@ def optimize_coefs_for_amplifications(
...
+    max_participations: Optional cap on the number of participations.
+    use_matrix_upper_bound: Whether to use the generic matrix-based upper bound
+      when the absolute strategy coefficients are not non-increasing. This is
+      computationally heavier than the default projected-coefficient upper
+      bound.
+
+  Returns:
+    The b-min-separated sensitivity of the implied strategy matrix $C$.
+  """
+  noising_coef = pad_coefs_to_n(noising_coef, n)
```
chai1any:
```text
I do not think we should pad `noising_coef` to length `n` before calling `inverse_coef(noising_coef, n)`.

This looks harmless mathematically, but it defeats the banded representation. `pad_coefs_to_n()` expands the vector to length `n`, `inverse_coef()` forwards that full length into `solve_banded()`, and `inverse_as_streaming_matrix()` uses `coef.shape[0]` as the effective bandwidth and state size. So a truly banded inverse ends up being treated like an `n`-band solve with many trailing zeros.

The result should be numerically the same, but the runtime and memory profile are worse than necessary. `inverse_coef(noising_coef, n)` already handles a short coefficient vector correctly, so I think this line should just be removed.

File: jax_privacy/matrix_factorization/toeplitz.py [hunk 3]
```diff
@@ -734,3 +734,167 @@ def optimize_coefs_for_amplifications(
...
+  strategy_coef = inverse_coef(noising_coef, n)
+  abs_strategy_coef = jnp.abs(strategy_coef)
+  is_nonincreasing = jnp.all(abs_strategy_coef[1:] <= abs_strategy_coef[:-1])
+
+  if not use_matrix_upper_bound:
+    coef_for_upper_bound = abs_strategy_coef
+    if not is_nonincreasing:
+      coef_for_upper_bound = jax.lax.cummax(abs_strategy_coef[::-1])[::-1]
+    return float(
```
chai1any: This eager `float(...)` conversion makes the reusable part of the implementation non-JAX and harder to share with the optimizer.

I think it would be cleaner if the core sensitivity helper returned a scalar `jax.Array`, and only the outermost caller converted to Python `float` when that is actually needed for display or reporting. That would make it much easier to reuse exactly the same safe sensitivity implementation inside `optimize_banded_inverse_toeplitz(...)`, and it would avoid splitting the logic into a JAX-compatible path in one place and an eager path in another.

examples/dpmf_strategy_optimization.py
```diff
@@ -104,6 +106,35 @@ def optimize_strategy(
...
+      # https://arxiv.org/pdf/2505.12128
+      weight_decay = 1.0
+      momentum = 0.0
+      workload_coef = toeplitz.multiply(
+          weight_decay**jnp.arange(n),
+          momentum**jnp.arange(n),
+          n=n,
+          skip_checks=True,
+      )
+      noising_coef = toeplitz.optimize_banded_inverse_toeplitz(
```
chai1any, line 111: There is an API mismatch here between the example CLI and the optimizer being called.

The example exposes `--objective={mean,max}` and threads `reduction_fn` into the existing `banded-toeplitz` path, but `optimize_banded_inverse_toeplitz(...)` has no `reduction_fn` parameter and its docstring explicitly says it optimizes mean per-query squared error. So if the user selects `--objective=max`, this branch still optimizes mean error internally and only changes the final reporting step.

I think this should be made explicit one way or the other:
1. either extend `optimize_banded_inverse_toeplitz(...)` with `reduction_fn`,
2. or reject unsupported objective choices in the example,
3. or document clearly that this branch is mean-only for now.

As written, the CLI can silently optimize one objective and report another.

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,167 @@ def optimize_coefs_for_amplifications(
...
+            noising_coef=coef,
+            n=n,
+            workload_coef=workload_coef,
+            skip_checks=True,
+        )
+    )
+    sens_squared = minsep_sensitivity_squared(
+        strategy_coef,
+        min_sep=min_sep,
+        max_participations=None,
```
chai1any: This also seems inconsistent with how the example evaluates the final result.

Inside the optimizer, the sensitivity term is computed with `max_participations=None`, but in `examples/dpmf_strategy_optimization.py` the final `compute_banded_inverse_sensitivity(...)` call uses `max_participations=participations`. That means the sensitivity used during optimization is not necessarily the same one used during evaluation.

I think `optimize_banded_inverse_toeplitz(...)` should accept `max_participations` and optimize against the same constraint that the example later reports, or the example should make it explicit that the current optimizer objective is only an approximation to the reported metric.

examples/dpmf_strategy_optimization.py
```diff
@@ -104,6 +106,35 @@ def optimize_strategy(
...
+          n=n,
+          min_sep=sep,
+          num_bands=sep,
+          weight_decay=weight_decay,
+          momentum=momentum,
+          max_optimizer_steps=10,
```
chai1any: `max_optimizer_steps=10` feels very low for a new optimization path, especially since the neighboring `banded-toeplitz` branch uses `max_optimizer_steps=1000`.

Unless there is evidence that this reliably converges in 10 L-BFGS iterations, I would either increase this substantially or make it configurable from flags. Right now it is hard to tell whether this branch is underperforming because of the factorization itself or simply because optimization is being stopped almost immediately.

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,167 @@ def optimize_coefs_for_amplifications(
...
+  Args:
+    num_bands: The number of coefficients to return.
+    weight_decay: The weight decay factor.
+    momentum: The momentum factor.
+
+  Returns:
+    The coefficients of the lower-triangular Toeplitz noising matrix $C^{-1}$.
+  """
+  k = jnp.arange(num_bands)
+  sqrt_coefs = jnp.cumprod(((k - 3 / 2) / k).at[0].set(1.0))
```
chai1any: This compact implementation is nice, but I think it really needs direct tests because the coefficient recurrence is subtle.

At minimum I would add:
1. a small regression test for `banded_inverse_square_root_noising_coefs()` on tiny `num_bands`,
2. a test that `compute_banded_inverse_sensitivity()` matches the closed-form Toeplitz value in a monotone case,
3. an optimizer regression test showing the returned coefficients are finite, normalized, and no worse than the initializer under the optimized objective.

The new API surface is mathematically nontrivial, so I do not think this should land with most of the new code path uncovered.

staticpayload (changes requested):
The biggest concern for me is the objective inside `optimize_banded_inverse_toeplitz()`. Right now it recovers `strategy_coef` with `solve_banded(...)` and then feeds that directly into `minsep_sensitivity_squared(..., skip_checks=True)`. My understanding is that the closed-form Toeplitz sensitivity path assumes a non-negative, non-increasing coefficient sequence, while the optimizer here is unconstrained. So during optimization it seems possible to leave the regime where that expression is justified, even though the objective still uses it directly. Since the post-hoc sensitivity helper already has the safer upper-bound logic based on `abs(...)` and the decreasing majorant, I think it would be better if the optimizer and the reporting path shared one JAX-native “sensitivity squared upper bound” helper.

The second issue I noticed is in `compute_banded_inverse_sensitivity()`, where `noising_coef` is padded to length `n` before calling `inverse_coef(noising_coef, n)`. That looks harmless at first, but it seems to give up the main benefit of the banded representation. Once padded, the solve effectively behaves like an `n`-band solve with lots of trailing zeros, rather than a genuinely banded solve. I think `inverse_coef(noising_coef, n)` should already handle the short-vector case correctly, so the padding step looks unnecessary and more expensive than it needs to be.

I also think there is a mismatch between the optimizer API and the example wiring. The example exposes `--objective={mean,max}`, but the new inverse-Toeplitz optimizer is currently mean-only. On top of that, the optimizer uses `max_participations=None`, while the example evaluates the final result with `max_participations=participations`. So the path can end up optimizing one thing and reporting another. Even if that is acceptable for a first version, I think it should either be made explicit in the example or the optimizer should be extended so the optimized and reported objectives match.

Last thing, I would feel much better about this with a few direct tests for the new helpers and optimizer. The new math is subtle enough that I think targeted regression tests would add a lot of confidence.

Overall, I do think this is a good addition. I just think the safety of the optimizer objective, the unnecessary padding before inversion, the example/API mismatch, and the lack of direct tests are worth addressing before merge.

npkalinin:
@staticpayload, thank you for taking the time to review the code.

Regarding the sensitivity computation, in my initial experiments, optimizing over the majorant decreasing positive sequence did not work; it did not converge. However, I have now tried to implement it carefully, and it seems to work well and reasonably fast. So, let's optimize over the safe sensitivity computation, which is in fact much better.

Previously, it was not a good idea to have much more than 10–30 iterations because it was slow and optimized the wrong objective. Now we can afford 1000 iterations, and I will fix it.

As for the choice of the reduction function, I had never tested it with the 'max' reduction before. I have now tried it, and it works just as well, so we can use it.

I added the tests. Please take a look at the new commits.

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,171 @@ def optimize_coefs_for_amplifications(
...
+  https://arxiv.org/pdf/2505.12128. The objective is the reduced per-query
+  squared error on the induced workload times the squared `min_sep`
+  sensitivity of the implied strategy matrix $C$.
+
+  Args:
+    n: The number of iterations that defines the workload.
+    min_sep: The minimum separation between contributions from the same user.
+    num_bands: The number of Toeplitz coefficients of the noising matrix to
+      optimize, including the diagonal.
+    noising_coef: Optional initialization for the noising coefficients. If not
```
ryan112358, line 872: noising_coef refers to the coefficients of C^{-1}.  I think it'd be more natural to parameterize this in terms of the "strategy_coef", I.e., the Toeplitz parameters of C.  This is what we do elsewhere in the library

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,171 @@ def optimize_coefs_for_amplifications(
...
+  )
+
+
+def optimize_banded_inverse_toeplitz(
+    n: int,
+    min_sep: int,
+    num_bands: int,
+    noising_coef: jax.Array | None = None,
+    max_participations: int | None = None,
+    weight_decay: float = 1.0,
```
ryan112358: Rather than parameterize in terms of a weight_decay and momentum, let's have an optional "workload" argument which is an array of size n referring to the toeplitz coefficients of A.  If None, it can default to all ones.  You can add a doctest showing how to build the momentum + weight decay workload.

ryan112358 (changes requested):
Looks great!  A few minor comments (in addition to @staticpayload's)

examples/dpmf_strategy_optimization.py
```diff
@@ -104,6 +106,36 @@ def optimize_strategy(
...
+          n=n,
+          min_sep=sep,
+          num_bands=sep,
+          weight_decay=weight_decay,
+          momentum=momentum,
+          max_optimizer_steps=1000,
+          reduction_fn=reduction_fn,
+      )
+      sensitivity_squared = toeplitz.compute_banded_inverse_sensitivity_squared(
+          n=n,
```
chai1any, line 120: Thanks for the follow-up changes — this addresses most of what I raised earlier.

I think there is still one remaining mismatch here: `optimize_banded_inverse_toeplitz(...)` now accepts `max_participations`, but this example branch still does not pass `participations` through. So the optimizer is still minimizing with `max_participations=None`, while the final `compute_banded_inverse_sensitivity_squared(...)` call below evaluates with `max_participations=participations`.

That means the optimized objective and the reported metric can still diverge in multi-participation settings. I think this call should also pass `max_participations=participations` so the example is optimizing and reporting the same quantity.

staticpayload:
Everything else looks good from my side!

examples/dpmf_strategy_optimization.py
npkalinin: Fixed it!

jax_privacy/matrix_factorization/toeplitz.py
npkalinin: I have removed weight_decay and momentum, now all functions accept an arbitrary workload, or jnp.ones(n) if None is provided

jax_privacy/matrix_factorization/toeplitz.py
npkalinin: I have added an option to initialize with strategy_coef, it wouldn't be a faithful initialization though, as the inverse of it might not be banded, then it will be clipped to num_bands

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,184 @@ def optimize_coefs_for_amplifications(
   coef = helper.optimize_bands(max_optimizer_steps=max_optimizer_steps)['coef']
   stddev = helper.required_stddev(coef)
   return coef, stddev
+
+
+def banded_inverse_square_root_noising_coefs(
```
ryan112358, line 741: Let's throw a jit decorator on this function, there are enough array copies below that it might be a bit innefficient otherwise.

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,184 @@ def optimize_coefs_for_amplifications(
...
+          max_participations=max_participations,
+      )
+      ** 2
+  )
+
+
+def optimize_banded_inverse_toeplitz(
+    n: int,
+    min_sep: int,
+    num_bands: int,
```
ryan112358, line 849: Let's make the remaining Args kw-only by adding a "*" here.

ryan112358 (approved):

jax_privacy/matrix_factorization/toeplitz.py
npkalinin: Done

jax_privacy/matrix_factorization/toeplitz.py
npkalinin: Done

ryan112358 (approved):

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,186 @@ def optimize_coefs_for_amplifications(
...
+  Args:
+    num_bands: The number of coefficients to return.
+    workload_coef: Optional Toeplitz coefficients of the workload.
+
+  Returns:
+    The coefficients of the lower-triangular Toeplitz noising matrix $C^{-1}$.
+  """
+  if workload_coef is None:
+    k = jnp.arange(num_bands)
+    return jnp.cumprod(((k - 3 / 2) / k).at[0].set(1.0))
```
ryan112358: Is this the same as optimal_max_error_noising_coefs(...)? If so we should probably use that? In any case, having some comments  relating these would be useful.

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,186 @@ def optimize_coefs_for_amplifications(
...
+
+  This function takes Toeplitz coefficients of a lower-triangular noising
+  matrix $C^{-1}$, computes the implied strategy coefficients for $C$, and then
+  estimates the min-separation sensitivity of $C$.
+
+  If the absolute strategy coefficients are non-increasing, this uses the
+  closed-form Toeplitz sensitivity computation in
+  `minsep_sensitivity_squared`. Otherwise, if `use_matrix_upper_bound` is
+  False, it projects the absolute strategy coefficients onto the smallest
+  non-increasing majorant and uses that to obtain an upper bound. If
```
ryan112358: Does it sometimes give a tighter bound? Always? We should give the user enough information to decide how to set this?

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,186 @@ def optimize_coefs_for_amplifications(
...
+    reduction_fn: A function that converts per query squared errors to a scalar.
+      Use jnp.mean to optimize mean-squared-error, jnp.max to optimize max
+      squared error, or lambda v: v[-1] to optimize last iterate squared error.
+      Defaults to jnp.mean.
+
+  Returns:
+    The optimized Toeplitz coefficients of the lower-triangular noising
+    matrix $C^{-1}$.
+  """
+  if workload_coef is None:
```
ryan112358, line 892: From Mutants:
> Changing this 1 line to
>   if (workload_coef is not None):
> does not cause any test exercising them to fail.
> 
> Add test cases that fail when code is changed to catch bugs.

Please fix.

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,186 @@ def optimize_coefs_for_amplifications(
...
+    workload_coef = jnp.ones(n)
+  else:
+    workload_coef = pad_coefs_to_n(workload_coef, n)
```
ryan112358: From Mutants:
> Even with these 2 lines (885-886) removed, the affected tests still pass.
> 
> Add test cases that fail when code is changed to catch bugs.

Please fix.

ryan112358 (changes requested):

jax_privacy/matrix_factorization/toeplitz.py
npkalinin: Apparently, optimal_max_error_noising_coefs does the same, I have switched to it, if the workload is not provided.

jax_privacy/matrix_factorization/toeplitz.py
npkalinin: use_matrix_upper_bound=True gives a tighter bound whenever the sequence is not decreasing. If the sequence is positive, then use_matrix_upper_bound=True computes the sensitivity exactly. If the sequence is positive and decreasing, use_matrix_upper_bound=False computes it exactly. However, since we optimize for the projected sensitivity, the two tend to be close to each other for BandInvMF.

jax_privacy/matrix_factorization/toeplitz.py
npkalinin: I think it would make sense to remove it, the workload if provided should be of size n, I will add the skip_checks flag for it

ryan112358 (approved):

jax_privacy/matrix_factorization/toeplitz.py
```diff
@@ -734,3 +734,198 @@ def optimize_coefs_for_amplifications(
...
+    min_sep: int,
+    num_bands: int,
+    *,
+    noising_coef: jax.Array | None = None,
+    strategy_coef: jax.Array | None = None,
+    workload_coef: jax.Array | None = None,
+    max_participations: int | None = None,
+    max_optimizer_steps: int = 1000,
+    reduction_fn: Callable[[jax.Array], jax.Array] = jnp.mean,
+    skip_checks: bool = False,
```
ryan112358: If this doesn't cause any problems, let's just hardcode skip_checks=False and not make it an argument.  It looks like the checks are only on static attributes of the array (like it's size), so we can probably just remove it in a follow-up cleanup PR.

PR #222, opened 2026-04-18
Follow up to PR #196: fix keras_api_example regression

staticpayload:
## Summary

Fixes the Keras MNIST example regression from #196 while keeping the example on the intended Poisson-sampled training path.

The example now uses a larger DP batch size (`512` instead of `128`), which improves utility enough to keep the existing `>0.85` train-accuracy assertion without weakening the test.

## Why this approach

`poisson_sampling_in_fit=True` is the right behavior for this example because the training batches now match the Poisson sampling assumption used by the accountant. The accuracy drop came from the real sampling change, not from a bug in the Poisson-backed `fit()` path: routing fixed-size batches through that path matched the default batching path closely.

Increasing the batch size is the smallest useful tuning change here. It keeps the example instructive, keeps accounting and training aligned, and avoids lowering the assertion.

## What changed

- Increase `batch_size` from `128` to `512`.
- Seed Keras initialization with `keras.utils.set_random_seed(0)`.
- Keep `batch_size` out of the DP `fit()` call so internal Poisson sampling is still used.
- Replace stale printed accuracy expectations with the thresholds the example actually asserts.

## Evidence

Local Poisson runs with `epsilon=1.1`, `epochs=5`:

| batch size | seed | train accuracy | val accuracy |
| --- | ---: | ---: | ---: |
| 256 | 0 | 0.8571 | 0.8011 |
| 512 | 0 | 0.8770 | 0.7888 |
| 512 | 1 | 0.8750 | 0.8739 |
| 1024 | 0 | 0.8820 | 0.8236 |

`512` is the best middle ground: enough margin for the current assertion, without jumping to the much larger `1024` batch.

## Validation

- `python examples/keras_api_example.py` passes locally with final DP train accuracy around `0.8770`.
- Non-DP sanity run with `batch_size=512`: train `0.9733`, val `0.9828`.
- GitHub CI is green for branch sync, lint, Linux tests, Windows tests, docs, heavy tests, and CLA.

### DIFF
--- a/docs/keras_api.rst
+++ b/docs/keras_api.rst
@@ -37,9 +37,11 @@ example below shows that.
 This section demonstrates how to integrate the Keras API into a typical
 Keras training workflow.
 
-The example below enables ``poisson_sampling_in_fit`` and passes training data
-to ``fit()`` as per-example arrays. In that setup, the DP Keras wrapper draws
-Poisson-sampled batches internally from those arrays.
+The example below uses the default DP-SGD path where Keras handles batching
+through the usual ``batch_size`` argument to ``fit()``. To opt into internal
+Poisson sampling from per-example arrays instead, set
+``DPKerasConfig.poisson_sampling_in_fit=True`` and pass those arrays directly to
+``fit()`` without a ``batch_size`` argument.
 
 .. literalinclude:: ../examples/keras_api_example.py
    :language: python
--- a/docs/keras_api.rst
+++ b/docs/keras_api.rst
@@ -37,11 +37,9 @@ example below shows that.
 This section demonstrates how to integrate the Keras API into a typical
 Keras training workflow.
 
-The example below uses the default DP-SGD path where Keras handles batching
-through the usual ``batch_size`` argument to ``fit()``. To opt into internal
-Poisson sampling from per-example arrays instead, set
-``DPKerasConfig.poisson_sampling_in_fit=True`` and pass those arrays directly to
-``fit()`` without a ``batch_size`` argument.
+The example below enables ``poisson_sampling_in_fit`` and passes training data
+to ``fit()`` as per-example arrays. In that setup, the DP Keras wrapper draws
+Poisson-sampled batches internally from those arrays.
 
 .. literalinclude:: ../examples/keras_api_example.py
    :language: python
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -57,6 +57,8 @@ Then on top of the core library the following backend-specific public high-level
 
 These APIs abstract some complexity and reduce the amount of code necessary to
 implement DP training at the cost of less flexibility. Currently, the only
-supported mechanism available when using the Keras API is DP-SGD with
-internally Poisson-sampled batches built from random-access per-example arrays
-(with accounting done using the same Poisson-sampling assumption).
+supported mechanism available when using the Keras API is DP-SGD. You can
+either use the standard Keras batching path, or opt into internally
+Poisson-sampled batches built from random-access per-example arrays by setting
+`poisson_sampling_in_fit=True` (with accounting done using the same
+Poisson-sampling assumption).
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -57,8 +57,6 @@ Then on top of the core library the following backend-specific public high-level
 
 These APIs abstract some complexity and reduce the amount of code necessary to
 implement DP training at the cost of less flexibility. Currently, the only
-supported mechanism available when using the Keras API is DP-SGD. You can
-either use the standard Keras batching path, or opt into internally
-Poisson-sampled batches built from random-access per-example arrays by setting
-`poisson_sampling_in_fit=True` (with accounting done using the same
-Poisson-sampling assumption).
+supported mechanism available when using the Keras API is DP-SGD with
+internally Poisson-sampled batches built from random-access per-example arrays
+(with accounting done using the same Poisson-sampling assumption).
--- a/examples/keras_api_example.py
+++ b/examples/keras_api_example.py
@@ -93,7 +93,6 @@ def main(_):
         batch_size=batch_size,
         train_steps=epochs * (train_size // batch_size),
         train_size=train_size,
-        poisson_sampling_in_fit=True,
         seed=0,
         gradient_accumulation_steps=1,
     )
@@ -102,8 +101,6 @@ def main(_):
         f"DP training:{epsilon=} {delta=} {clipping_norm=} {batch_size=} "
         f"{epochs=} {train_size=}"
     )
-    # This example opts into internal Poisson sampling from the per-example
-    # arrays passed to fit().
   else:
     print("Non-DP training")
   model.compile(
@@ -114,15 +111,16 @@ def main(_):
   fit_kwargs = dict(
       x=x_train,
       y=y_train,
+      batch_size=batch_size,
       epochs=epochs,
       validation_data=(x_test, y_test),
   )
-  if not dp:
-    fit_kwargs["batch_size"] = batch_size
   history = model.fit(**fit_kwargs)
   # [END example]
-  print("DP: expected train accuracy: ~96%, val accuracy: ~92%")
-  print("Non-DP: expected train accuracy: ~98%, val accuracy: ~98%")
+  if dp:
+    print("DP: expected train accuracy: >85% within 5 epochs")
+  else:
+    print("Non-DP: expected train accuracy: >95% within 5 epochs")
   final_accuracy = history.history["accuracy"][-1]
   if dp:
     assert (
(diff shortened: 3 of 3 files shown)


ryan112358 (commented):
Thanks for the fix @staticpayload .  Do you believe there is a bug in the poisson sampling codepath? Why do you think the utility is so much worse?

staticpayload:
I took another look at it, and at this point I don’t think I’ve found a bug in the Poisson path itself.

I reran the example locally with the same setup and only changed poisson_sampling_in_fit. On my side, the old fixed-batch DP path got to about 0.91 train accuracy after 5 epochs, and the Poisson-in-fit() path landed around 0.82. I also checked the wrapper code and the padding/masking logic, and nothing obvious looks wrong there.

My guess is that most of the drop is coming from the change in training regime, not from a correctness issue. The old example was effectively using fixed-size shuffled batches, while this path does true Poisson resampling. That makes optimization noisier: batch sizes vary, some examples get repeated or skipped within an epoch, and the same 5 epochs no longer means quite the same thing. Also, the fixed-batch DP path on my environment doesn’t hit the old ~96% / ~92% numbers either, so I think those expectations were already stale and the Poisson switch just pushed the example below the assertion.

Happy to dig further, but right now I’d describe this as a utility regression from switching the canonical example to Poisson sampling, not a confirmed bug in the Poisson implementation.

staticpayload:
Spent a bit more time on this and I still don’t think there’s a bug in the Poisson path itself.

What changed my mind was rerunning the example with the privacy noise turned off. Even without the DP noise, the fixed-batch version still ends up quite a bit higher than the Poisson version after 5 epochs. So a lot of the gap is already there before noise enters the picture.

That makes me think this is mostly about the training setup changing. The old path was using fixed-size shuffled batches, where you see the whole dataset every epoch. With true Poisson sampling, batch sizes move around and you don’t see every example each epoch, so 5 epochs just isn’t the same amount of optimization anymore.

There’s also a baseline issue here: the old fixed-batch path wasn’t really a like-for-like privacy baseline, because the Keras config calibrates noise under Poisson accounting assumptions by default.

So my current take is: the regression is real, but it looks more like a utility hit from switching the example to Poisson sampling than a bug in the Poisson implementation.

ryan112358 (approved):
Interesting, thanks for the investigation.  Your explanation seems plausible, but I am still surprised.  Did you rule out the following:

1) Average batch size (before padding) under Poisson matches
2) The way gradients are normalized is consistent across the methods? 

If so, this is an interesting finding.  Is there any existing research paper that has observed this specific phenomenon in other settings that you're aware of?

ryan112358:
> Also, the fixed-batch DP path on my environment doesn’t hit the old ~96% / ~92% numbers either, so I think those expectations were already stale and the Poisson switch just pushed the example below the assertion.

This statement also concerns me.  Can you figure out which PR introduced the regression on the default (non-poisson) codepath?

staticpayload:
I did one more pass on the two concrete questions and on the older fixed-batch numbers.

I still don't see evidence of a bug in the Poisson implementation itself.

On average batch size: yes, that checks out. In the `poisson_sampling_in_fit` path, the sampler uses `q = batch_size / train_size`, so the expected pre-padding batch size is 128 by construction for the MNIST example. I also wrote a small Python script to simulate the exact example settings (`train_size=59904`, `batch_size=128`, `steps_per_epoch=468`) over many epochs. The mean pre-padding batch size came out to `127.93`, with standard deviation `11.28`, which matches the Binomial expectation very closely.

On gradient normalization: I don't see a mismatch there either. Both paths normalize the clipped gradients by `dp_params.batch_size`, not by the realized batch size, so I don't think the utility gap is coming from a simple scaling inconsistency.

The bigger behavioral difference seems to be coverage per epoch. Under those same MNIST settings, Poisson sampling only touches about `63.2%` of distinct examples in an epoch on average, which is basically exactly `1 - (1 - q)^steps`. Fixed shuffled batching sees the whole dataset once per epoch. So even before adding DP noise, `5 epochs` is just not the same optimization budget under the two samplers.

On the older fixed-batch numbers: I still can't point to a repo PR after `v0.4.0` that introduced a regression in the default non-Poisson path. The old `~96% / ~92%` strings were already present in `5a35a95` (`v0.4.0`) as printed expectations. The stricter `>0.85` train assertion appears later in `5ad11bc`. I reran the fixed-batch example with deterministic initialization on `5ad11bc` and on this PR branch and got the same result on the current stack, so `#196` is definitely not the source of a fixed-batch regression.

So my best read at this point is:

- `#196` introduced the example failure by switching the canonical example to the harsher Poisson path
- I don't see a separate recent PR that regressed the default fixed-batch codepath
- the older expectation strings look stale relative to today's dependency stack, rather than being broken by one later repository change

On prior work: I don't know a paper that makes exactly this same "fixed number of epochs, Poisson vs shuffled" point for this example, but the closest things I found were:

- Chua et al., *Scalable DP-SGD: Shuffling vs. Poisson Subsampling* (NeurIPS 2024), which compares the samplers under matched privacy accounting and finds the big differences come from the privacy/accounting side rather than from Poisson being inherently broken
- the older SGD literature on without-replacement sampling, e.g. Shamir 2016, which notes that shuffled / without-replacement passes are often empirically better than with-replacement SGD over a fixed number of passes

So I think the surprising part here is real, but the evidence still points to "different training dynamics under a fixed epoch budget" rather than "bug in Poisson sampling."

ryan112358:
Thank you for the additional digging @staticpayload.  It would be great to get one more data point just to ensure consistency between the following two codepaths:

1) Run 1: The current default codepath (no poisson sampling)
2) Run 2: Use the Poisson sampling codepath, but change the strategy to one that does fixed size batches.  This can be accomplished with swapping out for CyclicPoissonSampling(sampling_prob=1, iterations=iterations, cycle_length=iterations/epochs)

If these two approaches match, it will give me additional confidence that there are no bugs.

arung54:
Thanks for the thorough investigation here.

Since the accounting is assuming Poisson sampling regardless of whether it is used or not, and the batch size is relatively small, it's not surprising switching to Poisson sampling creates a large drop in accuracy.

That being said, since this is an example, it seems like being instructive (e.g., showing how to enable Poisson sampling and keep training 1-to-1 with accounting) might be more important than achieving high accuracy. To that end, I think there are some alternatives to improve the accuracy when using DP that might be preferable to outright disabling Poisson sampling, e.g.:

- Increase epsilon significantly to reduce noise
- Increase batch size (at some cost in runtime)
- If the drop in accuracy is not catastrophic, lower the threshold for what we expect from DP training (maybe not ideal)

staticpayload:
I ran the extra comparison on the actual PR snapshot (`6f3efd3`) using the same MNIST setup and the same calibrated noise multiplier, and compared:

1. the current default path (`poisson_sampling_in_fit=False`)
2. the Poisson-backed `fit()` path with the sampler swapped to `CyclicPoissonSampling(sampling_prob=1.0, iterations=steps_per_epoch, cycle_length=steps_per_epoch)`

On the metric the example currently asserts (`history.history["accuracy"][-1]`), they match very closely:

- seed 0: default `0.9116`, Poisson-backed fixed-size `0.9109`
- seed 1: default `0.9129`, Poisson-backed fixed-size `0.9121`

For context, true Poisson at seed 0 was `0.8229`.

I also logged the internal batches. In the patched Poisson-backed run, every batch was exactly size `128` and the sampler covered all `59904` training examples each epoch. In the true Poisson run, the mean batch size was still about `128`, but distinct-example coverage was only about `63.3%` of the dataset per epoch.

So this makes me a lot more confident that there isn’t a bug in the Poisson `fit()` codepath itself. The utility drop seems to come from the actual switch in sampling regime, not from routing through the Poisson-backed implementation.

ryan112358 (approved):

ryan112358:
Thank you, that gives me more confidence in the correctness. As Arun suggests, let's try to tweak the hyper-parameters to try to close the gap: use a larger batch size in the example, and if necessary relax the accuracy assertion a bit. Then I think it'll be good to merge.

staticpayload:
Thanks, I pushed an update in `c0217ca`.

What changed:

- restored the example to the instructive Poisson path with `poisson_sampling_in_fit=True`
- increased the example `batch_size` from `128` to `512`
- kept the existing DP train accuracy assertion at `>0.85`
- added `keras.utils.set_random_seed(0)` so the example run is reproducible
- updated the surrounding docs text to match the Poisson-backed example again

Local validation on the PR branch:

- `batch_size=256`, seed 0: train history accuracy `0.8571`, val accuracy `0.8011`
- `batch_size=512`, seed 0: train history accuracy `0.8770`, val accuracy `0.7888`
- `batch_size=512`, seed 1: train history accuracy `0.8750`, val accuracy `0.8739`
- `batch_size=1024`, seed 0: train history accuracy `0.8820`, val accuracy `0.8236`

I chose `512` as the middle ground because it clears the current assertion with reasonable margin while avoiding the larger per-example batch of `1024`. I also ran the actual `examples/keras_api_example.py` after the patch, and it passed with final train accuracy around `0.8770`, so I did not relax the assertion.

ryan112358 (approved):

PR #238, opened 2026-05-15
Fix multi-participation sensitivity calculation for banded Toeplitz strategies

npkalinin:
This PR fixes an inconsistency in the multi-participation sensitivity calculation in `examples/dpmf_strategy_optimization.py.
`

For the ` banded-toeplitz`, `banded-sqrt`, and `normalized-banded-toeplitz` methods, `strategy_coefficients` were computed with length `sep` and then passed to `toeplitz.minsep_sensitivity_squared`. Since that function infers the matrix size from the provided coefficients, it treated the number of iterations as `sep` rather than `n`.

As a result, the squared sensitivity was underestimated by a factor of `participations`, making the reported RMSE inconsistent with the `blt`, `dense`, and `banded-inverse` methods.

The fix pads the strategy coefficients to length `n` before calling `toeplitz.minsep_sensitivity_squared`, ensuring that the sensitivity is computed for the intended number of iterations.

### DIFF
--- a/examples/dpmf_strategy_optimization.py
+++ b/examples/dpmf_strategy_optimization.py
@@ -100,6 +100,7 @@ def optimize_strategy(
       strategy_coef = toeplitz.optimize_banded_toeplitz(
           n, bands=sep, max_optimizer_steps=1000, reduction_fn=reduction_fn
       )
+      strategy_coef = jnp.pad(strategy_coef, (0, n - len(strategy_coef)))
       sensitivity_squared = toeplitz.minsep_sensitivity_squared(
           strategy_coef, min_sep=sep, max_participations=participations
       )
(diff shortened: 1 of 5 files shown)


examples/dpmf_strategy_optimization.py
```diff
@@ -100,6 +100,7 @@ def optimize_strategy(
       strategy_coef = toeplitz.optimize_banded_toeplitz(
           n, bands=sep, max_optimizer_steps=1000, reduction_fn=reduction_fn
       )
+      strategy_coef = jnp.pad(strategy_coef, (0, n - len(strategy_coef)))
       sensitivity_squared = toeplitz.minsep_sensitivity_squared(
```
ryan112358, line 103: Can you just pass n to minsep_sensitivity_squared here instead?

This is a subtle bug that indicates to me we should maybe make n a required argument as well, although that is maybe out of scope for this bug-fix PR.

ryan112358 (changes requested):

examples/dpmf_strategy_optimization.py
npkalinin: Done! I made n a required parameter and propagated it in all functions that used it.

ryan112358 (approved):

ryan112358 (approved):

PR #255, opened 2026-06-09
Guard against ZeroDivisionError for noise_multiplier=0 with use_zcdp=True

r1cksync:
## Bug

`experimental.accounting._validate_args` permits `noise_multiplier >= 0`, and the continuous branch handles `GaussianDpEvent(0)` gracefully, but the zCDP branch computes `0.5 / noise_multiplier**2` and raises an opaque `ZeroDivisionError` on the validated-as-legal input `noise_multiplier=0`:

```python
>>> dpsgd_event(0.0, 1, sampling_prob=1.0, use_zcdp=True)
ZeroDivisionError: float division by zero
```

## Fix

Guard the division and return `ZCDpEvent(inf)` (ρ → ∞, i.e. no privacy) for `noise_multiplier=0`, matching the degenerate semantics already implied by the continuous branch. No `noise_multiplier > 0` result changes. Adds a regression test covering both branches.

## Testing
`tests/experimental/accounting_test.py` — pass. The new `test_dpsgd_event_zero_noise_does_not_crash` errors with `ZeroDivisionError` on the pre-fix code; existing zCDP structure tests are unaffected.

### DIFF
--- a/jax_privacy/accounting.py
+++ b/jax_privacy/accounting.py
@@ -109,7 +109,8 @@ def _gaussian_event(
 ) -> dp_accounting.DpEvent:
   """Returns the Gaussian or zCDP DpEvent for the given noise multiplier."""
   if use_zcdp:
-    return dp_accounting.ZCDpEvent(0.5 / noise_multiplier**2)
+    rho = math.inf if noise_multiplier == 0 else 0.5 / noise_multiplier**2
+    return dp_accounting.ZCDpEvent(rho)
   else:
     return dp_accounting.GaussianDpEvent(noise_multiplier)
 
--- a/tests/accounting_test.py
+++ b/tests/accounting_test.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import functools
+import math
 
 from absl.testing import absltest
 from absl.testing import parameterized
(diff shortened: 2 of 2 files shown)


jax_privacy/accounting.py
```diff
@@ -109,7 +109,13 @@ def _gaussian_event(
 ) -> dp_accounting.DpEvent:
   """Returns the Gaussian or zCDP DpEvent for the given noise multiplier."""
   if use_zcdp:
-    return dp_accounting.ZCDpEvent(0.5 / noise_multiplier**2)
+    # noise_multiplier == 0 is permitted by the validators (which only reject
+    # negative values) and denotes infinite privacy loss (zCDP rho -> inf). Guard
+    # the 0.5 / noise_multiplier**2 division so this branch matches the Gaussian
+    # branch, which already handles GaussianDpEvent(0) gracefully, instead of
+    # raising an opaque ZeroDivisionError on a validated-as-legal input.
+    rho = math.inf if noise_multiplier == 0 else 0.5 / noise_multiplier**2
```
ryan112358, line 112: Thanks for the bug fix, I think this is pretty self explanatory so please delete the inline comment

ryan112358 (changes requested):

jax_privacy/accounting.py
r1cksync: Done — removed the inline comment. Thanks for the review!

r1cksync:
Rebased onto the latest main. Since `jax_privacy/experimental/accounting.py` was promoted to the top-level `jax_privacy/accounting.py` (and the test to `tests/accounting_test.py`), I moved the fix and its regression test to the new locations — the bug was still present there. Also dropped the inline comment per your earlier note. The conflict is resolved and `tests/accounting_test.py` passes. Thanks for the review!

ryan112358 (approved):

Issue #234, opened 2026-05-12
Inconsistent `calculated_noise_multiplier` when using `gradient_accumulation_steps` in Gemma3 LoRA example


SunnierLee:
Hi JAX Privacy team,

Thank you for maintaining this excellent library. I am currently exploring the `examples/dp_sgd_keras_gemma3_lora_finetuning_samsum.ipynb` example and have encountered a discrepancy regarding how `train_steps` and `gradient_accumulation_steps` affect the noise multiplier calculation.

### Context
In the notebook, it is stated:
> "Also note that `gradient_accumulation_steps` is provided as a separate parameter. JAX Privacy takes it into account to calculate the real number of optimization steps during the training."

Based on this, I understood that if I provide the total number of *micro-batch* steps as `train_steps` (i.e., `EPOCHS * (TRAIN_SIZE // BATCH_SIZE)`), the library should internally divide this by `gradient_accumulation_steps` to determine the actual number of *optimizer updates* (macro-steps) for the privacy accounting.

### Observed Behavior
I ran two experiments with the same effective batch size ($1024$) and privacy budget ($\epsilon=4.0, \delta=2e-05$), but different physical batch sizes and gradient accumulation steps. The dataset size is $14,732$.

**Configuration 1:**
*   `epoch`: 3
*   `batch_size`: 4
*   `gradient_accumulation_steps`: 256
*   `effective_batch_size`: $4 \times 256 = 1024$
*   `train_steps` (passed): $3 \times (14732 // 4) = 11,049$ (Total micro-steps)
*   **Output:**
    ```text
    Calculating noise multiplier for: self.epsilon=4.0, self.delta=2e-05, self.effective_batch_size=1024, self.train_steps=11049, self.train_size=14732.
    Finished calculating noise multiplier: calculated_noise_multiplier=7.6917...
    ```

**Configuration 2:**
*   `epoch`: 3
*   `batch_size`: 16
*   `gradient_accumulation_steps`: 64
*   `effective_batch_size`: $16 \times 64 = 1024$
*   `train_steps` (passed): $3 \times (14732 // 16) = 2,760$ (Total micro-steps)
*   **Output:**
    ```text
    Calculating noise multiplier for: self.epsilon=4.0, self.delta=2e-05, self.effective_batch_size=1024, self.train_steps=2760, self.train_size=14732.
    Finished calculating noise multiplier: calculated_noise_multiplier=3.9159...
    ```

### The Issue
[...]

PR #267, opened 2026-06-14
docs(keras_api): correct train_steps formula to account for gradient accumulation

Osamaali313:
## Summary

The `DPKerasConfig.train_steps` docstring contradicts itself, which is the root cause of the confusion reported in #234.

- Line: *"`train_steps`: The number of training steps (**optimizer update steps**)"*
- But the formula it gives is `epochs * (train_size // batch_size)`, which is the number of **(micro-)batches**, not optimizer updates.

These only agree when `gradient_accumulation_steps == 1`. With gradient accumulation, one optimizer update consumes `gradient_accumulation_steps` batches, so the documented formula is `gradient_accumulation_steps`× too large. A user who follows it passes an inflated step count to the accountant — exactly the discrepancy in #234, where two runs with the same effective batch size but different `gradient_accumulation_steps` produced different noise multipliers (7.69 vs 3.92).

## Fix

Documentation only. Correct the epochs formula to use the **effective** batch size and clarify the iterator case:

```
epochs * (train_size // effective_batch_size),
  where effective_batch_size = batch_size * gradient_accumulation_steps
```

## Why docs, not the accounting code

The privacy accountant is **correct**: it uses `train_steps` as the number of privatized optimizer updates (`iterations=self.train_steps`), which is the right composition count for DP accounting. Changing the accountant to divide `train_steps` by `gradient_accumulation_steps` would *under-count* iterations and **silently weaken the privacy guarantee** — so the bug is in the docstring formula, not the math. This change carries no behavioral or privacy risk.

Addresses #234.

### DIFF
--- a/jax_privacy/keras_api.py
+++ b/jax_privacy/keras_api.py
@@ -99,10 +99,17 @@ class DPKerasConfig:
         to memory constraints, therefore gradient accumulation technique is very
         useful during DP training.
       train_steps: The number of training steps (optimizer update steps). If you
-        try to train the model for more steps, it will fail. If you train by
-        epochs, then it is epochs * (train_size // batch_size). If you train
-        while the dataset iterator is not over then it is the length of the
-        dataset iterator.
+        try to train the model for more steps, it will fail. Because one
+        optimizer update consumes `gradient_accumulation_steps` batches, this is
+        the number of optimizer updates, not the number of (micro-)batches. If
+        you train by epochs, then it is
+        epochs * (train_size // effective_batch_size), where
+        effective_batch_size = batch_size * gradient_accumulation_steps (this
+        reduces to epochs * (train_size // batch_size) only when
+        gradient_accumulation_steps == 1). If you train while the dataset
+        iterator is not over, then it is the number of optimizer updates the
+        iterator produces, i.e. its (micro-)batch length //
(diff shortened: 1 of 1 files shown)


Entropy-rgb (approved):

RamSaw (changes requested):
thank you very much! great change, lgtm.

one thing: could you reformat it this way so that our internal formatting does not complain
```
      train_steps: The number of training steps (optimizer update steps). If you
        try to train the model for more steps, it will fail. Because one
        optimizer update consumes `gradient_accumulation_steps` batches, this is
        the number of optimizer updates, not the number of (micro-)batches. If
        you train by epochs, then it is epochs * (train_size //
        effective_batch_size), where effective_batch_size = batch_size *
        gradient_accumulation_steps (this reduces to epochs * (train_size //
        batch_size) only when gradient_accumulation_steps == 1). If you train
        while the dataset iterator is not over, then it is the number of
        optimizer updates the iterator produces, i.e. its (micro-)batch length
        // gradient_accumulation_steps.
```

Osamaali313:
Done — rewrapped the `train_steps` docstring exactly as you laid out (pushed in 9fc4e2f). Thanks for the review and the precise formatting!

RamSaw (approved):
Thank you very much!

PR #268, opened 2026-06-14
Add functionality for per-layer clipping

notimeforcaution:
Per-layer, translates to per-leaf in the PyTree based implementation. Currently, only a singular value of max L2 Norm per leaf is supported. Upcoming change will allow the users to provide max L2-norm per leaf as a PyTree, where the L2 Norm PyTree will be a "prefix tree" of the Gradient PyTree.


notimeforcaution:
Closing this in-interest of https://github.com/google-deepmind/jax_privacy/pull/269, in-order to avoid merging using main branch of local repo.

PR #269, opened 2026-06-14
Add functionality for per-layer clipping

notimeforcaution:
Per-layer, translates to per-leaf in the PyTree based implementation. Currently, only a singular value of max L2 Norm per leaf is supported. Upcoming change will allow the users to provide max L2-norm per leaf as a PyTree, where the L2 Norm PyTree will be a "prefix tree" of the Gradient PyTree.

### DIFF
--- a/jax_privacy/clipping.py
+++ b/jax_privacy/clipping.py
@@ -18,6 +18,7 @@ import collections
 from collections.abc import Sequence
 import dataclasses
 import functools
+import math
 import numbers
 from typing import Any, Callable, TypeAlias
 
@@ -28,9 +29,11 @@ import jax.numpy as jnp
 import optax
 from optax import microbatching
 
-
 PyTree: TypeAlias = chex.ArrayTree
 AuxiliaryOutput = collections.namedtuple('Aux', ['values', 'grad_norms', 'aux'])
+ClippedTreeAndNorm = collections.namedtuple(
+    '_ClippedTreeAndNorm', ['clipped', 'norm']
+)
 _REPLACE_SPECIAL = dp_accounting.NeighboringRelation.REPLACE_SPECIAL
 
 
@@ -81,15 +84,25 @@ class BoundedSensitivityCallable:
 
 def clip_pytree(
     pytree: PyTree,
-    clip_norm: float,
+    clip_norm: float | PyTree,
     rescale_to_unit_norm: bool = False,
     nan_safe: bool = True,
     return_zero: bool = False,
 ):
-  """Clips a PyTree of jax arrays based on its global L2 norm.
+  """Clips a PyTree of jax arrays.
 
-  Calculates the global L2 norm of the input PyTree. If the norm exceeds
-  `clip_norm`, the PyTree is scaled down to have norm equal to `clip_norm`.
+    The clipping behavior is determined by the type of `clip_norm`, like so:
+
+    - If `clip_norm` is a Scalar (Global Clipping): Calculates the global L2
+      norm of the input PyTree; & if the norm exceeds `clip_norm`, the entire
+      PyTree is scaled down.
+    - If `clip_norm` is a PyTree (Per-Layer Clipping): Calculates the L2 norm
+      for each individual leaf; & if a leaf's norm exceeds its corresponding
+      threshold in `clip_norm`, that specific leaf is scaled down. The
+      `clip_norm` PyTree structure must be a prefix of the input `pytree`.
+      Where `clip_norm` terminates early, its leaf values automatically
+      broadcast down to match all corresponding sub-leaves of `pytree`.
+
   If `rescale_to_unit_norm` is True, the PyTree is additionally scaled by
   `1.0 / clip_norm` (resulting in a norm of at most 1.0 no matter
   what clip_norm is). Handles cases where the original norm is zero,
@@ -117,7 +130,9 @@ def clip_pytree(
 
   Args:
     pytree: The PyTree of arrays to clip.
-    clip_norm: The maximum L2 norm allowed.
+    clip_norm: The maximum L2 norm allowed. Can be a single float or a PyTree;
+      depending on whether Global / "Per-Layer" clipping is to be done (see
+      details above).
     rescale_to_unit_norm: If True, the output PyTree's norm is rescaled by `1.0
       / clip_norm` after potential clipping. If False, the output PyTree has
       norm at most `clip_norm`.
@@ -132,7 +147,30 @@ def clip_pytree(
   Returns:
     A tuple `(clipped_pytree, original_l2_norm)`, where `clipped_pytree` is the
     processed PyTree and `original_l2_norm` is the L2 norm of the input PyTree.
+    In case of per-layer clipping the original_l2_norm is a PyTree containing
+    l2 norms of each leaf.
   """
+  # -- Per-layer Norm --
+  if not jnp.isscalar(clip_norm):
+    recursive_clip = functools.partial(
+        clip_pytree,
+        rescale_to_unit_norm=rescale_to_unit_norm,
+        nan_safe=nan_safe,
+        return_zero=return_zero,
+    )
+    results = jax.tree.map(
+        lambda c_norm, p_tree: ClippedTreeAndNorm(
+            *recursive_clip(p_tree, c_norm)
+        ),
+        clip_norm,
+        pytree,
+    )
+    is_leaf = lambda x: isinstance(x, ClippedTreeAndNorm)
+    clipped = jax.tree.map(lambda x: x.clipped, results, is_leaf=is_leaf)
+    norms = jax.tree.map(lambda x: x.norm, results, is_leaf=is_leaf)
+    return ClippedTreeAndNorm(clipped=clipped, norm=norms)
+
+  # -- Global Norm --
   if isinstance(clip_norm, numbers.Real) and clip_norm < 0:
     raise ValueError(f'clip_norm must be non-negative, got {clip_norm=}.')
 
@@ -363,7 +401,7 @@ def clipped_fun(
     *,
     batch_argnums: int | Sequence[int] = 0,
     keep_batch_dim: bool = True,
-    l2_clip_norm: float = 1.0,
+    l2_clip_norm: float | PyTree = 1.0,
     rescale_to_unit_norm: bool = False,
     normalize_by: float = 1.0,
     return_norms: bool = False,
@@ -408,7 +446,8 @@ def clipped_fun(
     keep_batch_dim: If True, batch inputs will be passed to `fun` with a leading
       batch axis of size 1.  If False, this size 1 axis will be dropped
       (reducing the rank of the batch args by 1 before passing to `fun`).
-    l2_clip_norm: The maximum L2 norm allowed.
+    l2_clip_norm: The maximum L2 norm allowed. Can be a single float or a
+      PyTree; depending on whether Global / "Per-Layer" clipping is to be done.
     rescale_to_unit_norm: If True, the output PyTree's norm is rescaled by `1.0
       / clip_norm` after potential clipping. If False, the output PyTree has
       norm at most `clip_norm`.
@@ -464,6 +503,11 @@ def clipped_fun(
       raise ValueError(
           'normalize_by is not compatible with grid_scale. Normalization '
           'should be applied after noise addition.'
+      )
+    if not jnp.isscalar(l2_clip_norm):
+      raise ValueError(
+          'Per-layer PyTree clipping is not compatible with grid scale.'
+          'l2_clip_norm must be a Real scalar.'
       )
     _check_x64_enabled()
 
@@ -548,8 +592,16 @@ def clipped_fun(
 
   if grid_scale is not None:
     norm_bound = float(grid_scale)
+  elif rescale_to_unit_norm:
+    # If every leaf is rescaled to unit norm, the worst-case global norm is
+    # the square root of the total number of leaves in the PyTree.
+    num_leaves = len(jax.tree.leaves(l2_clip_norm))
+    norm_bound = math.sqrt(num_leaves) / normalize_by
   else:
-    norm_bound = (1.0 if rescale_to_unit_norm else l2_clip_norm) / normalize_by
+    norm_bound = (
+        sum(leaf**2 for leaf in jax.tree.leaves(l2_clip_norm)) ** 0.5
+    ) / normalize_by
+
   if keep_batch_dim:
     clipped_fn = _with_extra_batch_axis(clipped_fn, batch_argnums)
   callable_has_aux = has_aux or return_norms
@@ -583,7 +635,7 @@ def clipped_grad(
     argnums: int | Sequence[int] = 0,
     has_aux: bool = False,
     *,
-    l2_clip_norm: float,
+    l2_clip_norm: float | PyTree,
     rescale_to_unit_norm: bool = False,
     normalize_by: float = 1.0,
     batch_argnums: int | Sequence[int] = 1,
@@ -643,17 +695,18 @@ def clipped_grad(
 
   Formal Guarantees:
     For the gradient output:
-      The L2 sensitivity of the returned function with respect to the batch
-      arguments (specified by `batch_argnums`) under add/remove or zero-out
-      differential privacy definitions is guaranteed to be 1.0 if
-      `rescale_to_unit_norm` is True. Otherwise, the sensitivity is
-      `l2_clip_norm`. Under replace-one DP, the sensitivity is doubled
-      (2.0 or 2 * `l2_clip_norm`).
+      The L2 sensitivity of the returned callable is a complex function of
+      the input parameters (``l2_clip_norm``, ``rescale_to_unit_norm``,
+      ``grid_scale``, per-layer clipping settings, etc.) and may change as
+      new features are added. Rather than reasoning about the sensitivity
+      from the parameters, callers should query the returned callable's
+      ``.sensitivity()`` method (or equivalently its ``.l2_norm_bound``
+      attribute for add-or-remove-one DP) directly.
     All auxiliary outputs (aux, values, grad_norms) are per-example. This
-      function guarantees that per-example outputs only depend the data for the
-      same example. This allows maximum flexibility for the caller to aggregate
-      these as desired (possibly with a DP mean, median, quantile, or histogram
-      mechanism).
+      function guarantees that per-example outputs only depend on the data for
+      the same example. This allows maximum flexibility for the caller to
+      aggregate these as desired (possibly with a DP mean, median, quantile,
+      or histogram mechanism).
 
   Args:
     fun: The function to be differentiated, which should return a scalar loss
@@ -666,7 +719,13 @@ def clipped_grad(
       Exercise caution when using this as no DP sensitivity guarantees are
       provided for the auxiliary data.
     l2_clip_norm: The maximum L2 norm for each per-example gradient. Gradients
-      with a norm larger than this value will be scaled down.
+      with a norm larger than this value will be scaled down. Can be a single
+      float or a PyTree. If Scalar: It does global clipping across all
+      gradients. If PyTree: It does per-layer clipping and contains l2 norm
+      associated with each leaf of the PyTre. The `clip_norm` PyTree structure
+      must be a prefix of the input `pytree`. Where `clip_norm` terminates
+      early, its leaf values automatically broadcast down to match all
+      corresponding sub-leaves of `pytree`.
     rescale_to_unit_norm: If True, clipped gradients are rescaled by `1.0 /
       l2_clip_norm`. This ensures the sensitivity is 1.0. If False, they are
       only scaled down if their norm exceeds `l2_clip_norm`, resulting in a
@@ -689,7 +748,11 @@ def clipped_grad(
     return_values: If True, the transformed function will also return the
       per-example values, before clipping.
     return_grad_norms: If True, the transformed function will also return the
-      per-example gradient norms, before clipping.
+      per-example gradient norms, before clipping. If `l2_clip_norm` is a
+      scalar, it returns a single global norm for the entire gradient PyTree. If
+      `l2_clip_norm` is a PyTree, it returns a matching PyTree structure where
+      each leaf contains the L2 norm of the corresponding subtree in the
+      gradient PyTree.
     pre_clipping_transform: An optional function to apply to the per-example
       gradients before clipping. The function should consume the gradient pytree
       for a single example and returned a new pytree (possibly with different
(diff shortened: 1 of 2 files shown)


notimeforcaution:
Results of lint-based checks:

```
✦ ~/Documents/jax_privacy add-per-layer-clipping*
venv ❯ pytype tests/clipping_test.py
Computing dependencies
Analyzing 1 sources with 0 local dependencies
ninja: Entering directory `.pytype'
[1/1] check clipping_test
Leaving directory '.pytype'
Success: no errors found

✦ ~/Documents/jax_privacy add-per-layer-clipping*
venv ❯ flake8 jax_privacy/**.py      

✦ ~/Documents/jax_privacy add-per-layer-clipping*
venv ❯ pylint jax_privacy/**.py

✦ ~/Documents/jax_privacy add-per-layer-clipping* 6s
venv ❯ pytype jax_privacy/**.py
Computing dependencies
Analyzing 12 sources with 10 local dependencies
ninja: Entering directory `.pytype'
[16/16] check jax_privacy.__init__
Leaving directory '.pytype'
Success: no errors found

✦ ~/Documents/jax_privacy add-per-layer-clipping* 22s
venv ❯ flake8 tests/clipping_test.py

✦ ~/Documents/jax_privacy add-per-layer-clipping*
venv ❯ pylint tests/clipping_test.py
************* Module clipping_test
tests/clipping_test.py:1:0: C0114: Missing module docstring (missing-module-docstring)
tests/clipping_test.py:205:10: W0212: Access to a protected member _adjusted_clip_norm of a client class (protected-access)
tests/clipping_test.py:213:6: W0212: Access to a protected member _adjusted_clip_norm of a client class (protected-access)

✦ ~/Documents/jax_privacy add-per-layer-clipping*
venv ❯ pytype tests/clipping_test.py
Computing dependencies
Analyzing 1 sources with 0 local dependencies
ninja: Entering directory `.pytype'
[1/1] check clipping_test
Leaving directory '.pytype'
Success: no errors found
```

There are some lint errors in unrelated lines in tests/ directory. In order to keep changes isolated I am avoiding to change these lines in this PR. I will send some follow-up changes to fix these lint errors.

jax_privacy/clipping.py
```diff
@@ -584,6 +621,7 @@ def clipped_grad(
     has_aux: bool = False,
     *,
     l2_clip_norm: float,
+    per_layer: bool = False,
```
ryan112358: Per your proposal document, I think it's actually better to not add new arguments here, but just allow l2_clip_norm to be a pytree of floats instead.  I thought this is what you initially proposed

ryan112358 (changes requested):

jax_privacy/clipping.py
notimeforcaution: My thinking earlier was, to have 'per layer' arg, and let l2_clip_norm both. Then, let the latter either be a Real Number (in case we wish to apply same clip norm value on each leaf) or a PyTree (in case we wish to apply separate clip norm values on different leaves). Now, upon looking closely, this does feel overengineered.

Your suggestion (IIUC) is:
- Removing per_layer
- Allowing l2_clip_norm to accept a PyTree or a RealNumber. In case its PyTree we do a per layer clipping, else we do a Global Clipping.

That actually looks very good to me, and helps us remove a potential redundancy in args. I will make these changes, and update. If you think my understanding of your suggestion is incorrect, feel free to suggest that as well.

jax_privacy/clipping.py
```diff
@@ -139,16 +145,60 @@ def clip_pytree(
...
-  clip_norm = jnp.maximum(clip_norm, 0.0)
-  l2_norm = optax.tree.norm(pytree)
-  scale = jnp.minimum(1.0, clip_norm / l2_norm)
-  if rescale_to_unit_norm:
-    scale = jax.lax.select(clip_norm > 0, scale / clip_norm, 1 / l2_norm)
-  # If l2_norm is 0 or nan, set scale to 0.0.
-  scale = jnp.nan_to_num(scale, nan=0.0, posinf=0.0)
-  clipped = jax.tree.map(lambda x: jnp.astype(scale, x.dtype) * x, pytree)
+
+  def _calculate_scale(l2_norm, c_norm):
```
ryan112358: This is significantly more complicated than the original.  Can you try to simplify it?  I think you could set it up as a 2-layer recursive call.  If the l2_clip_norm is not a numeric type / jax.Tracer -> call jax.tree.map ove rthe leaves of clip norm.  Otherwise, keep the current implementation.  This would be a much lighter weight change I'd be more comfortable accepting

jax_privacy/clipping.py
```diff
@@ -522,16 +579,6 @@ def clipped_fun_one_group(*args, is_padding_example, **kwargs):
 
     clipped_values, aux, norms = microbatched_vmap_fun(*args, **kwargs)
     if keep_batch_dim:
-      # If keep_batch_dim is True, we artificially added a dimension of size 1
```
ryan112358, line 524: Avoid changing unrelated code

jax_privacy/clipping.py
```diff
@@ -465,6 +516,14 @@ def clipped_fun(
           'normalize_by is not compatible with grid_scale. Normalization '
           'should be applied after noise addition.'
       )
+    is_global_clip = jax.tree_util.treedef_is_leaf(
```
ryan112358: jax.tree_util is an old pattern I think, it's more common to use jax.tree.is_leaf, although I'm not familiar with this symbol.  Can you just check if l2_clip_norm is not a numeric type?

jax_privacy/clipping.py
```diff
@@ -465,6 +516,14 @@ def clipped_fun(
...
+        jax.tree_util.tree_structure(l2_clip_norm)
+    )
+    if not is_global_clip:
```
ryan112358: I recently added _validate.py, maybe this check can go there as "is_numeric" or something

jax_privacy/clipping.py
```diff
@@ -741,9 +811,17 @@ def grad_fn(*args, **kwargs):
     value_and_aux, grad = value_and_grad_fn(*args, **kwargs)
     result = pre_clipping_transform(grad)
     if has_aux or return_values or return_grad_norms:
+      if return_grad_norms:
+        if isinstance(l2_clip_norm, numbers.Real):
+          grad_norms_val = optax.tree.norm(grad)
+        else:
+          grad_norms_val = jax.tree.map(jnp.linalg.norm, grad)
```
ryan112358: This gives you the per-leaf norms, but if l2_clip_norm is a prefix of the parameter tree, then you might want the grad norms to match that structure.

jax_privacy/clipping.py
```diff
@@ -741,9 +811,17 @@ def grad_fn(*args, **kwargs):
     value_and_aux, grad = value_and_grad_fn(*args, **kwargs)
     result = pre_clipping_transform(grad)
     if has_aux or return_values or return_grad_norms:
+      if return_grad_norms:
```
ryan112358: I think rather than all this nesting, just do this

grad_norms = jax.tree.map(lambda _, g: optax.tree.norm(g), l2_clip_norm, grad)

Down below you can still have the check if "return_grad_norms" is True.  If it is False, then JIT will compile-away this dead code, so you don't need the extra level of nested conditional.

jax_privacy/clipping.py
```diff
@@ -666,7 +734,9 @@ def clipped_grad(
       Exercise caution when using this as no DP sensitivity guarantees are
       provided for the auxiliary data.
     l2_clip_norm: The maximum L2 norm for each per-example gradient. Gradients
-      with a norm larger than this value will be scaled down.
+      with a norm larger than this value will be scaled down. Can be a single
```
ryan112358, line 722: The conventions of return_grad_norm also changed.  Be sure to add a sentence to the doc about that.

tests/clipping_test.py
```diff
@@ -54,7 +54,10 @@ def test_clip_pytree_output_dtype_matches_input(self, pytree, **kwargs):
     pytree = optax.tree.random_like(jax.random.key(0), pytree)
     clipped, _ = clipping.clip_pytree(pytree, **kwargs)
 
-    jitted = jax.jit(clipping.clip_pytree, static_argnums=(2, 3))
```
ryan112358, line 57: This seems unrelated

jax_privacy/clipping.py
```diff
@@ -741,9 +811,17 @@ def grad_fn(*args, **kwargs):
...
+      else:
+        grad_norms_val = None
+
       aux = AuxiliaryOutput(
           values=value_and_aux[0] if return_values else None,
-          grad_norms=optax.tree.norm(grad) if return_grad_norms else None,
+          grad_norms=grad_norms_val,
           aux=value_and_aux[1] if has_aux else None,
       )
       return result, aux
```
ryan112358, line 815: You will also have to updat ethe "l2_norm_bound" of the BoundedSensitivityCallable.  Now, it should be the l2 norm of the leaves of l2_clip_norm, I guess l2_norm_bound=optax.tree.norm(l2_clip_norm) is the right value

tests/clipping_test.py
```diff
@@ -158,6 +161,98 @@ def test_clip_pytree_with_inf(self):
...
+        'extra_key': invalid_clip_extra_key,
+        'type_mismatch': invalid_clip_type_mismatch,
+    }
+
+    for name, invalid_clip in mismatch_cases.items():
+      with self.subTest(name=name):
+        with self.assertRaises(ValueError):
+          clipping.clip_pytree(pytree, invalid_clip)
+
+  def test_clip_pytree_with_prefix_tree_clip_norm(self):
```
ryan112358: See if you can merge some of these tests into a smaller number of parameterized tests.  You could sweep l2_clip_norm over a few values covering scalar, pytree-matching-leaves, and pytree-prefix.  Maybe it can even be added to an existing test, not sure.

tests/clipping_test.py
```diff
@@ -434,6 +539,84 @@ def fun(batched_data):
     self.assertEqual(aux_3d.shape, (15, 4, 9))
     self.assertEqual(aux_4d.shape, (15, 4, 9, 10))
 
+  def test_return_norms_per_layer_structure(self):
```
ryan112358: If you do the parameterized test above, just fold this check in there rather than creating a completely new test.  chex.assert_tree_structure_equal(l2_clip_norm, grad_norms) should suffice (or something like it).

ryan112358 (changes requested):

jax_privacy/clipping.py
notimeforcaution: Yes, completely agreed!

---

Side Note: This PR is a WIP, after changes to the initial proposed API design. In this version, it is purely functional (and not production-ready) code. I wanted to clean changes, and optimize wherever feasible before sending for review. However, I saw that it was marked as Open (and not Draft); and I am sorry for the same.

I have marked this as Draft now, and am working on addressing comments, and cleaning the logic. I will send it for review just as that is done.

tests/clipping_test.py
notimeforcaution: Yes, to be deleted. It was originally there because in earlier iteration I had added "per_layer" as a separate argument, and this order was stale. Considering that we are not adding new arguments now, this can be deleted.

jax_privacy/clipping.py
notimeforcaution: Done.

Used `jnp.isscalar(clip_norm)`, which returns true when clip_norm is a tracer or scalar; works best here!

jax_privacy/clipping.py
notimeforcaution: jnp.isscalar does the exact same thing (https://docs.jax.dev/en/latest/_autosummary/jax.numpy.isscalar.html#jax-numpy-isscalar), so going ahead with that. We probably won't require a custom check.

jax_privacy/clipping.py
```diff
@@ -132,6 +145,24 @@ def clip_pytree(
...
+          leaf_clip_norm,
+          rescale_to_unit_norm=rescale_to_unit_norm,
+          nan_safe=nan_safe,
+          return_zero=return_zero,
+      )
+
+    results = jax.tree.map(recursive_clip, clip_norm, pytree)
+    clipped = jax.tree.map(lambda _, r: r[0], clip_norm, results)
+    leaf_norms = jax.tree.map(lambda _, r: r[1], clip_norm, results)
+    return clipped, optax.tree.norm(leaf_norms)
```
ryan112358: I actually thing the contract of this function should be to return the clipped pytree along with the leaf norms

jax_privacy/clipping.py
```diff
@@ -132,6 +145,24 @@ def clip_pytree(
     A tuple `(clipped_pytree, original_l2_norm)`, where `clipped_pytree` is the
     processed PyTree and `original_l2_norm` is the L2 norm of the input PyTree.
   """
+  # -- Per-layer Norm --
+  if not jnp.isscalar(clip_norm):
+
+    def recursive_clip(leaf_clip_norm, leaf_tree):
```
ryan112358: Consider using functools.partial here

jax_privacy/clipping.py
```diff
@@ -547,8 +584,17 @@ def clipped_fun_one_group(*args, is_padding_example, **kwargs):
 
   if grid_scale is not None:
     norm_bound = float(grid_scale)
+  elif rescale_to_unit_norm:
+    # If every leaf is rescaled to unit norm, the worst-case global norm is
+    # the square root of the total number of leaves in the PyTree.
+    num_leaves = len(jax.tree_util.tree_leaves(l2_clip_norm))
```
ryan112358: prefer jax.tree.leaves over jax.tree_util.tree_leaves

jax_privacy/clipping.py
```diff
@@ -547,8 +584,17 @@ def clipped_fun_one_group(*args, is_padding_example, **kwargs):
...
+    norm_bound = math.sqrt(num_leaves) / normalize_by
   else:
-    norm_bound = (1.0 if rescale_to_unit_norm else l2_clip_norm) / normalize_by
+    norm_bound = optax.tree.norm(l2_clip_norm) / normalize_by
+    # Cast to a float only when norm_bound is not a JAX Tracer.
+    if not isinstance(norm_bound, jax.core.Tracer):
```
ryan112358: Ah this is interesting.  I guess optax.tree.norm will convert the output to a jax array even if the input is define w.r.t. primitives.  You'll probably have to write your own _primitive_tree_norm helper function using a combination of jax.tree.flatten and np.linalg.norm (numpy not jax).

Actually np.linalg.norm will assume the input is always primitive.  Technically the l2_clip_norm could be either primitive or jax scalars, so you need to avoid np.linalg.norm and manually do sum(leaf**2 for leaf in jax.tree.flatten(l2_clip_norm)).  Since it's a 1-liner, you might not need a helper function.

jax_privacy/clipping.py
```diff
@@ -607,6 +653,15 @@ def clipped_grad(
   gradient to have a norm of at most `l2_clip_norm`, and finally sums these
   clipped gradients.
 
+  The clipping behavior is determined by the type of `l2_clip_norm`:
```
ryan112358: I'd remove this documentation from here, and just expand the documentatoin of the l2_clip_norm Arg below

tests/clipping_test.py
```diff
@@ -434,6 +440,118 @@ def fun(batched_data):
     self.assertEqual(aux_3d.shape, (15, 4, 9))
     self.assertEqual(aux_4d.shape, (15, 4, 9, 10))
 
+    @parameterized.named_parameters(
```
ryan112358: I believe there are some property-based tests in this file that you can expand to increase test coverage as well . It would probably be a bit cleaner than this huge parameterized test

ryan112358 (changes requested):
Overall it's looking quite clean (except the test), nice work!

jax_privacy/clipping.py
ryan112358: sum(leaf**2 for leaf in jax.tree.flatten(l2_clip_norm)[0])**0.5

jax_privacy/clipping.py
```diff
@@ -132,6 +145,24 @@ def clip_pytree(
...
+      return clip_pytree(
+          leaf_tree,
+          leaf_clip_norm,
+          rescale_to_unit_norm=rescale_to_unit_norm,
+          nan_safe=nan_safe,
+          return_zero=return_zero,
+      )
+
+    results = jax.tree.map(recursive_clip, clip_norm, pytree)
+    clipped = jax.tree.map(lambda _, r: r[0], clip_norm, results)
```
ryan112358: This could be written as jax.tree.transpose(jax.tree.structure(pytree), None, results)

jax_privacy/clipping.py
```diff
@@ -131,7 +144,28 @@ def clip_pytree(
...
+        nan_safe=nan_safe,
+        return_zero=return_zero,
+    )
+    results = jax.tree.map(
+        lambda clip_norm, pytree: recursive_clip(pytree, clip_norm),
+        clip_norm,
+        pytree,
+    )
+    is_tuple = lambda x: isinstance(x, tuple)
+    clipped = jax.tree.map(lambda x: x[0], results, is_leaf=is_tuple)
```
ryan112358: The "is_tuple" check is brittle, because the pytree itself could contain a real tuple that is not a leaf.  Would jax.tree.transpose work here?

jax_privacy/clipping.py
```diff
@@ -142,11 +176,11 @@ def clip_pytree(
   l2_norm = optax.tree.norm(pytree)
   scale = jnp.minimum(1.0, clip_norm / l2_norm)
   if rescale_to_unit_norm:
-    scale = jax.lax.select(clip_norm > 0, scale / clip_norm, 1 / l2_norm)
+    scale = jnp.where(clip_norm > 0, scale / clip_norm, 1 / l2_norm)
```
ryan112358: This seems unrelated to the current PR, why change?

tests/clipping_test.py
```diff
@@ -100,40 +129,89 @@ def test_clip_pytree_clip_norm_zero_rescale(self):
...
       *cartesian_product(
-          clip_norm=[0.0, 1.0, 2.0, jnp.inf],
+          clip_norm=[
+              0.0,
+              1.0,
+              2.0,
+              jnp.inf,
+              {
+                  'a': np.array(jnp.inf),
+                  'b': np.array(jnp.nan),
```
ryan112358: Nan in clipping norm is not an edge case I've thought about, what's the intended behavior there?

tests/clipping_test.py
```diff
@@ -100,40 +129,89 @@ def test_clip_pytree_clip_norm_zero_rescale(self):
...
+                  'c': np.array(jnp.inf),
+              },
+          ],
           return_zero=[False, True],
           rescale_to_unit_norm=[False, True],
       )
   )
   def test_nan_safety(self, **kwargs):
-    pytree = jnp.array([3.0, 0.0, 1.0, jnp.nan, 2.0])
```
ryan112358, line 131: Just make this a two-element dict pytree by default for all tests and avoid the if / else

tests/clipping_test.py
```diff
@@ -100,40 +129,89 @@ def test_clip_pytree_clip_norm_zero_rescale(self):
     chex.assert_trees_all_close(clipped, expected)
 
   @parameterized.parameters(
-      *cartesian_product(pytree=PYTREE_STRUCTS, clip_norm=[0.0, 1.0, jnp.inf])
+      *cartesian_product(
```
ryan112358: Try to keep the existing parameterized annotation on one line, here and elsewhere , e.g., clip_norm=[0.0, 1.0, jnp.inf, {'a': 1, 'b': 2}]

tests/clipping_test.py
```diff
@@ -61,22 +61,51 @@ def test_clip_pytree_output_dtype_matches_input(self, pytree, **kwargs):
...
+              },
+          ],
           rescale_to_unit_norm=[True, False],
           nan_safe=[True, False],
           return_zero=[True, False],
       )
   )
   def test_clip_pytree_has_bounded_norm(self, pytree, **kwargs):
+    # Expect a ValueError when pytree is a leaf, and clip_norm is a PyTree.
+    if isinstance(kwargs['clip_norm'], dict) and not isinstance(pytree, dict):
```
ryan112358: These if / else are too complicated, please simplify

tests/clipping_test.py
```diff
@@ -61,22 +61,51 @@ def test_clip_pytree_output_dtype_matches_input(self, pytree, **kwargs):
   @parameterized.parameters(
       *cartesian_product(
           pytree=PYTREE_STRUCTS,
-          clip_norm=[0.0, 1.0, 2.0, jnp.inf],
+          clip_norm=[
```
ryan112358: Please keep clip_norm= to one line

jax_privacy/clipping.py
ryan112358: Perhaps the safest thing to do would be to define a _ClippedTreeAndNorm namedtuple.  so you can check against that specifically.  But given that this is a tricky point, I would say you should create a new parameterized tests of clip pytree that hits it with different combinations of parameters / l2_clip_norm structures to ensure it runs without error (primarily) and productes correct results (secondarily)

jax_privacy/clipping.py
```diff
@@ -131,7 +144,37 @@ def clip_pytree(
...
+        '_ClippedTreeAndNorm', ['clipped', 'norm']
+    )
+    results = jax.tree.map(
+        lambda c_norm, p_tree: _ClippedTreeAndNorm(
+            *recursive_clip(p_tree, c_norm)
+        ),
+        clip_norm,
+        pytree,
+    )
+    is_clipped_tree_and_norm_node = lambda x: isinstance(x, _ClippedTreeAndNorm)
```
ryan112358: Let's just call this "is_leaf", then the two tree.maps should each fit on one line below.

jax_privacy/clipping.py
```diff
@@ -131,7 +144,37 @@ def clip_pytree(
...
   """
+  # -- Per-layer Norm --
+  if not jnp.isscalar(clip_norm):
+    recursive_clip = functools.partial(
+        clip_pytree,
+        rescale_to_unit_norm=rescale_to_unit_norm,
+        nan_safe=nan_safe,
+        return_zero=return_zero,
+    )
+    _ClippedTreeAndNorm = collections.namedtuple(
```
ryan112358: I'd move this definition outside of the scope of this function.  Rather than map the raw results of recursive_clip, just have clip_pytree always return this named tuple

jax_privacy/clipping.py
```diff
@@ -547,8 +596,16 @@ def clipped_fun_one_group(*args, is_padding_example, **kwargs):
...
     norm_bound = float(grid_scale)
+  elif rescale_to_unit_norm:
+    # If every leaf is rescaled to unit norm, the worst-case global norm is
+    # the square root of the total number of leaves in the PyTree.
+    num_leaves = len(jax.tree.leaves(l2_clip_norm))
+    norm_bound = math.sqrt(num_leaves) / normalize_by
   else:
-    norm_bound = (1.0 if rescale_to_unit_norm else l2_clip_norm) / normalize_by
+    norm_bound = (
+        sum(leaf**2 for leaf in jax.tree.flatten(l2_clip_norm)[0]) ** 0.5
```
ryan112358: I guess jax.tree.leaves is slightly cleaner than jax.tree.flatten here

tests/clipping_test.py
```diff
@@ -435,5 +543,85 @@ def fun(batched_data):
     self.assertEqual(aux_4d.shape, (15, 4, 9, 10))
 
 
+class ClippedFunPerLayerTest(parameterized.TestCase):
+
+  @parameterized.named_parameters(
```
ryan112358: Same here

ryan112358 (changes requested):

ryan112358 (approved):

PR #270, opened 2026-06-15
Fix epsilon_from_gdp to return a valid high-confidence lower bound

vvv214:
## Summary

`CanaryScoreAuditor.epsilon_from_gdp` should return a high-confidence lower bound. The previous implementation used

```python
np.abs(isf(FPR_ub) - ppf(FNR_ub))
```

as a shortcut for the reverse `D', D` direction. With Clopper-Pearson upper confidence bounds, that shortcut can be too optimistic: a negative forward-direction `mu` is not automatically valid reverse-direction evidence, because the reverse direction needs its own error-rate bounds and threshold frontier.

This PR computes the two directions explicitly:

- one-sided GDP `mu` on the original `D, D'` score frontier;
- one-sided GDP `mu` on the swapped `D', D` score frontier;
- one Bonferroni budget over both frontiers' FPR/FNR confidence bounds;
- one final conversion from `max(mu_forward, mu_reverse, 0)` to `(epsilon, delta)`.

## Tests

Added coverage for null, small-sample null, forward-separated, and reverse-separated scores. Against the previous implementation:

| scenario | previous implementation | this PR |
|---|---:|---:|
| null scores | fails (`epsilon ~= 3.54`) | passes (`epsilon = 0`) |
| small-sample null scores | fails (`epsilon ~= 5.92`) | passes (`epsilon = 0`) |
| forward-separated scores | passes | passes |
| reverse-separated scores | fails (`epsilon = 0`) | passes (`epsilon ~= 15.81`) |

Also checked locally:

```text
uvx pyink --check --diff jax_privacy/auditing.py tests/auditing_test.py
git diff --check
python3 -m py_compile jax_privacy/auditing.py tests/auditing_test.py
```

### DIFF
--- a/tests/auditing_test.py
+++ b/tests/auditing_test.py
@@ -591,14 +591,14 @@ class CanaryScoreAuditorTest(parameterized.TestCase):
       self.assertGreaterEqual(max_accuracy, expected_max_accuracy)
 
   @parameterized.product(
-      mu=(-3.0, -1.0, -0.3, 0.1, 0.3, 1.0, 3.0),
+      mu=(-3.0, -1.0, 0.1, 0.3, 1.0, 3.0),
       out_samples_ratio=(0.5, 1.0, 1.5),
   )
   def test_epsilon_from_gdp_tight(self, mu, out_samples_ratio):
     rng = np.random.default_rng(seed=0xBAD5EED)
     significance = 0.1
     delta = 5e-2
-    in_samples = 1_500_000 if abs(mu) == 0.3 else 500_000
+    in_samples = 500_000
     out_samples = int(in_samples * out_samples_ratio)
     in_canary_scores = rng.normal(mu, 1, in_samples)
     out_canary_scores = rng.normal(0, 1, out_samples)
--- a/tests/auditing_test.py
+++ b/tests/auditing_test.py
@@ -591,63 +591,27 @@ class CanaryScoreAuditorTest(parameterized.TestCase):
       self.assertGreaterEqual(max_accuracy, expected_max_accuracy)
 
   @parameterized.product(
-      mu=(0.1, 0.3, 1.0, 3.0),
+      mu=(-3.0, -1.0, -0.3, 0.1, 0.3, 1.0, 3.0),
       out_samples_ratio=(0.5, 1.0, 1.5),
   )
   def test_epsilon_from_gdp_tight(self, mu, out_samples_ratio):
     rng = np.random.default_rng(seed=0xBAD5EED)
     significance = 0.1
     delta = 5e-2
-    in_samples = 500_000
+    in_samples = 1_500_000 if abs(mu) == 0.3 else 500_000
     out_samples = int(in_samples * out_samples_ratio)
     in_canary_scores = rng.normal(mu, 1, in_samples)
     out_canary_scores = rng.normal(0, 1, out_samples)
     auditor = auditing.CanaryScoreAuditor(in_canary_scores, out_canary_scores)
     eps = auditor.epsilon_from_gdp(significance, delta)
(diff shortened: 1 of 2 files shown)


tests/auditing_test.py
```diff
@@ -607,6 +607,42 @@ def test_epsilon_from_gdp_tight(self, mu, out_samples_ratio):
...
+    delta = 1e-5
+    m = 5000
+    in_canary_scores = rng.normal(0, 1, m)
+    out_canary_scores = rng.normal(0, 1, m)
+    auditor = auditing.CanaryScoreAuditor(in_canary_scores, out_canary_scores)
+
+    eps = auditor.epsilon_from_gdp(significance, delta)
+    self.assertLessEqual(eps, 0.1)
+
+  def test_epsilon_from_gdp_separated_is_positive(self):
```
galenandrew-google: This test is less specific than the earlier test [test_epsilon_from_gdp_tight]. Please remove it.

tests/auditing_test.py
```diff
@@ -607,6 +607,42 @@ def test_epsilon_from_gdp_tight(self, mu, out_samples_ratio):
     true_eps = dp_accounting.get_epsilon_gaussian(1 / mu, delta)
     np.testing.assert_allclose(eps, true_eps, rtol=0.05)
 
+  def test_epsilon_from_gdp_null_is_zero(self):
```
galenandrew-google: If I understand correctly, the change in this PR means that if the in_canary_scores have *smaller* values on average than out_canary_scores, the returned epsilon is zero. If so, we should test that case like:

in_canary_scores = rng.normal(-1, 1, m)
out_canary_scores = rng.normal(0, 1, m)

I would suggest that instead of this test and test_epsilon_from_gdp_small_sample_null_is_zero below, we have a single parameterized product test that looks at
(large sample, small sample) x (mu 0, mu negative)

galenandrew-google (commented):
Thank you for your change. The code looks good but I believe the tests can be improved.

tests/auditing_test.py
vvv214: removed

tests/auditing_test.py
vvv214: no; the returned epsilon is not zero. this PR's effect is roughly like swapping in_canary_scores and out_canary_scores, so it's above zero

I have reorged the testcases: the negative mu case is added to [test_epsilon_from_gdp_tight] parameterized.product

I kept [test_epsilon_from_gdp_null_is_zero], as when mu is zero, epsilon should be near zero; original code returns a positive value (e.g., epsilon ~= 3.54)

galenandrew-google:
I'm afraid this change introduces algorithmic complexity and especially inefficiency that is not justified by its utility. In particular, _get_tn_fn_counts is a very expensive operation, which is why it is called only once in the initializer. The additional complexity of writing _mu_from_gdp_counts and calling it twice is also not justified. All of this would just be to handle the case where a user accidentally swaps in_scores and out_scores. I see your point that the use of abs is not correct due to the bounds. I would prefer the solution to not attempt to account for this error. Simply remove the abs and return a mu of zero in case the maximum is negative.

vvv214:
> I'm afraid this change introduces algorithmic complexity and especially inefficiency that is not justified by its utility. In particular, _get_tn_fn_counts is a very expensive operation, which is why it is called only once in the initializer. The additional complexity of writing _mu_from_gdp_counts and calling it twice is also not justified. All of this would just be to handle the case where a user accidentally swaps in_scores and out_scores. I see your point that the use of abs is not correct due to the bounds. I would prefer the solution to not attempt to account for this error. Simply remove the abs and return a mu of zero in case the maximum is negative.

ok. will do that version

galenandrew-google (approved):
Thanks for the simpler change.

PR #271, opened 2026-06-15
Validate explicit noise_multiplier at the effective batch size

vvv214:
## Summary

`update_with_calibrated_noise_multiplier` and `_validate_params` must account at the **same** sampling rate, but the validation path's `sampling_prob` dropped the `gradient_accumulation_steps` factor that calibration includes:

- calibration: `sampling_prob = effective_batch_size / train_size` (= `batch_size * gradient_accumulation_steps / train_size`)
- validation (before): `sampling_prob = batch_size / train_size`

With `gradient_accumulation_steps > 1` the validation under-accounts (smaller sampling rate → smaller epsilon) and can accept a user-supplied `noise_multiplier` that actually exceeds the privacy budget.

## Fix

One line: make validation use `effective_batch_size / train_size`, identical to the calibration path (same `sampling_prob`, same `iterations=train_steps`). This is privacy-conservative — it can only ever reject more — and a no-op when `gradient_accumulation_steps == 1`.

## Test

`test_validate_noise_multiplier_uses_effective_batch_size`: with `gradient_accumulation_steps=8` (batch_size=10, train_steps=100, train_size=1000, target epsilon=1.1, delta=1e-5), `noise_multiplier=1.0` gives epsilon ≈ 0.72 at the physical batch size (which the old check wrongly accepted) but ≈ 5.6 at the effective batch size, so it must be rejected. Fails on the old code (no raise), passes after.

### DIFF
--- a/jax_privacy/keras_api.py
+++ b/jax_privacy/keras_api.py
@@ -227,7 +227,7 @@ class DPKerasConfig:
     if self.noise_multiplier is not None:
       _validate.positive(noise_multiplier=self.noise_multiplier)
       try:
-        sampling_prob = self.batch_size / self.train_size
+        sampling_prob = self.effective_batch_size / self.train_size
         event = accounting.dpsgd_event(
             noise_multiplier=self.noise_multiplier,
             iterations=self.train_steps,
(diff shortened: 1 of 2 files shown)


vvv214:
Closing this PR: `import/copybara` persistently errored here ("An error happened while migrating the change", no `cl/` link) across multiple fresh pushes, and I couldn't determine why. Reopened the identical change as #287, which is importing cleanly. Superseded by #287.

PR #272, opened 2026-06-15
Warn when accounting assumes Poisson but poisson_sampling_in_fit=False

vvv214:
## Summary
The Keras DP-SGD accountant always assumes Poisson subsampling amplification, but `poisson_sampling_in_fit` defaults to `False`. In that default configuration the reported `(epsilon, delta)` silently assumes amplification the data path does not provide. This PR adds a warning (no numeric change).

## The issue
`accounting.dpsgd_event` (used by both `update_with_calibrated_noise_multiplier` and the `_validate_params` epsilon check) always wraps the Gaussian in a `PoissonSampledDpEvent`. But with `poisson_sampling_in_fit=False` (the default), the wrapped `fit()` consumes the user's own pre-batched data — typically shuffled, fixed-size minibatches — which does **not** satisfy Poisson subsampling. The amplified epsilon is then reported with no indication that the guarantee depends on a sampling scheme the default pipeline doesn't implement.

## The change
Emit a `UserWarning` from `DPKerasConfig` when `poisson_sampling_in_fit=False`, telling the user to either set it to `True` or ensure their input pipeline performs Poisson subsampling. No accounting/noise/sensitivity behavior changes.

```
The reported (epsilon, delta) privacy guarantee assumes Poisson subsampling
amplification, but poisson_sampling_in_fit=False. The default fit() path
consumes your pre-batched (typically shuffled, fixed-size) data and does NOT
perform Poisson sampling, so the computed epsilon may understate the true
privacy loss. Either set poisson_sampling_in_fit=True, or ensure your input
pipeline performs Poisson subsampling before passing data to fit().
```

## Notes / open questions
- This is intentionally the minimal, conservative change (a warning). A deeper fix would be to account with a non-amplified event when `poisson_sampling_in_fit=False` — happy to take direction on whether you'd prefer that.
- Placement is in `_validate_params` so it fires on config construction. It currently warns on every default config; if you'd rather it fire once at calibration/`fit` time, easy to move.

## Verification
- New `test_warns_when_poisson_sampling_in_fit_disabled` and `test_no_warning_when_poisson_sampling_in_fit_enabled`.
- Full `keras_api_test.py` (37 tests) passes (no `filterwarnings=error` in the project config).

### DIFF
--- a/jax_privacy/keras_api.py
+++ b/jax_privacy/keras_api.py
@@ -48,6 +48,7 @@ import inspect
 import math
 import types
 from typing import Any
+import warnings
 
 import chex
 import dp_accounting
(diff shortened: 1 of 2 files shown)


vvv214:
closed as i think this might be a bigger issue and should be addressed more systematically

PR #273, opened 2026-06-15
Fix BMinSepSampling(min_sep=1) crash and correct docstrings

vvv214:
## Summary
Fixes a crash in `BMinSepSampling` for `min_sep=1`, plus two documentation corrections (one of them privacy-relevant).

## The bug
`BMinSepSampling.batch_iterator` crashes for `min_sep=1` with the default `warm_start=True`. The warm-start path builds history via `_independent_partition(size, min_sep - 1, ...)`, i.e. 0 groups, which hits `rng.multinomial(n, [])`:

```
ValueError: pvals must have at least 1 dimension and the last dimension of pvals must be greater than 0.
```

`min_sep=1` is a valid configuration (`__post_init__` only requires `min_sep > 0`) and is the natural "no separation = plain Poisson" setting.

## The fix
Skip the warm-start when `min_sep == 1` (the history must be empty when there is no separation constraint), so the sampler reduces to plain Poisson(p).

## Documentation fixes (same change set)
- **`noise_addition.py`** — `matrix_factorization_privatizer` documents the realized noise covariance as `noising_matrix.T @ noising_matrix`. The privatizer emits `M @ z` per step, so the covariance is `noising_matrix @ noising_matrix.T`. Verified empirically: a Monte-Carlo sample covariance of the privatizer output matches `M @ M.T` (max abs deviation ≈ 0.006) and clearly not `M.T @ M` (≈ 0.54). Corrected both occurrences (no code change).
- **`batch_selection.py`** — `CyclicPoissonSampling` referenced a nonexistent `even_partition` attribute (corrected to `partition_type = EQUAL_SPLIT`), and the `truncated_batch_size` doc said "If True" though the field is `int | None` (corrected to "If set").

## Verification
- New `test_b_min_sep_sampling_min_sep_one_warm_start`: constructs `BMinSepSampling(..., min_sep=1)` and iterates without error (crashes on current code); asserts per-example inclusion frequency ≈ `sampling_prob` (reduces to Poisson).
- New `test_b_min_sep_sampling_min_sep_one_poisson_frequency`.
- Full `batch_selection_test.py` (84 tests) passes.

### DIFF
--- a/tests/batch_selection_test.py
+++ b/tests/batch_selection_test.py
@@ -417,6 +417,41 @@ class BatchSelectionTest(parameterized.TestCase):
         self.assertGreaterEqual(len(batches[0]), 400)
         self.assertLessEqual(len(batches[0]), 600)
 
+  def test_b_min_sep_sampling_min_sep_one_warm_start(self):
+    """min_sep=1 with default warm_start reduces to plain Poisson(p)."""
+    sampling_prob = 0.01
+    iterations = 5
+    num_examples = 1000
+    strategy = batch_selection.BMinSepSampling(
+        sampling_prob=sampling_prob,
+        iterations=iterations,
+        min_sep=1,
+    )
+    # warm_start defaults to True; with min_sep == 1 (b == 1) there is no
+    # separation constraint and the warm-start history must be empty.
+    self.assertTrue(strategy.warm_start)
+    batches = list(strategy.batch_iterator(num_examples, rng=0))
+    self.assertLen(batches, iterations)
+    _check_element_range(batches, num_examples)
+    _check_signed_indices(batches)
+
+  def test_b_min_sep_sampling_min_sep_one_poisson_frequency(self):
+    """Over many iterations, min_sep=1 gives per-example frequency ~= p."""
+    sampling_prob = 0.1
+    iterations = 2000
+    num_examples = 200
+    strategy = batch_selection.BMinSepSampling(
+        sampling_prob=sampling_prob,
+        iterations=iterations,
+        min_sep=1,
+    )
+    batches = list(strategy.batch_iterator(num_examples, rng=0))
+    self.assertLen(batches, iterations)
+    mean_inclusion_freq = sum(len(batch) for batch in batches) / (
+        iterations * num_examples
+    )
+    self.assertAlmostEqual(mean_inclusion_freq, sampling_prob, delta=0.003)
+
   def test_user_selection_strategy(self):
     """Tests for UserSelectionStrategy."""
     base_strategy = batch_selection.CyclicPoissonSampling(
--- a/jax_privacy/batch_selection.py
+++ b/jax_privacy/batch_selection.py
@@ -251,7 +251,7 @@ class CyclicPoissonSampling(BatchSelectionStrategy):
     - With truncation, if > truncated_batch_size examples appear in a batch
       under the previous guarantee, then we select truncated_batch_size of
       them uniformly at random and discard the rest.
-    - If even_partition = True, num_examples % cycle_length examples are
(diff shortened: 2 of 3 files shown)


tests/batch_selection_test.py
```diff
@@ -417,6 +417,53 @@ def test_b_min_sep_sampling_prob_one(
...
+    # warm_start defaults to True; with min_sep == 1 (b == 1) there is no
+    # separation constraint and the warm-start history must be empty.
+    self.assertTrue(strategy.warm_start)
+    batches = list(strategy.batch_iterator(num_examples, rng=0))
+    self.assertLen(batches, iterations)
+    _check_element_range(batches, num_examples)
+    _check_signed_indices(batches)
+    # b=1 means no separation constraint: any pair of batches may overlap, and
+    # _check_b_min_sep_property is trivially satisfied for min_sep=1.
+    _check_b_min_sep_property(batches, 1)
```
arung54: I'd delete this check, it's a no_op for min_sep = 1 (the inner for loop in the helper loops over an empty range)

tests/batch_selection_test.py
```diff
@@ -417,6 +417,53 @@ def test_b_min_sep_sampling_prob_one(
...
+        min_sep=1,
+    )
+    batches = list(strategy.batch_iterator(num_examples, rng=0))
+    self.assertLen(batches, iterations)
+    counts = collections.defaultdict(int)
+    for batch in batches:
+      for x in batch:
+        counts[x] += 1
+    freqs = np.array([counts[i] for i in range(num_examples)]) / iterations
+    self.assertAlmostEqual(float(np.mean(freqs)), sampling_prob, delta=0.02)
```
arung54: Since you're only checking the mean there's no need to keep the individual counts. This check could be simplified e.g. by replacing `np.mean(freqs)` with ` sum([len(b) for b in batches]) / iterations / num_examples`

arung54: The standard deviation of this value is about .0005, the delta here can probably be tightened a bit (e.g. a delta of .003 would still mean 6 standard deviations, so a ~1 in billion chance of failure)

tests/batch_selection_test.py
```diff
@@ -417,6 +417,53 @@ def test_b_min_sep_sampling_prob_one(
...
+    # separation constraint and the warm-start history must be empty.
+    self.assertTrue(strategy.warm_start)
+    batches = list(strategy.batch_iterator(num_examples, rng=0))
+    self.assertLen(batches, iterations)
+    _check_element_range(batches, num_examples)
+    _check_signed_indices(batches)
+    # b=1 means no separation constraint: any pair of batches may overlap, and
+    # _check_b_min_sep_property is trivially satisfied for min_sep=1.
+    _check_b_min_sep_property(batches, 1)
+    # With min_sep=1 the sampler reduces to plain Poisson(p): each example is
```
arung54: Since test_b_min_sep_sampling_min_sep_one_poisson_frequency is already checking this (with much higher sample count), I'd remove testing of inclusion_freq here to simplify / avoid redudancy between tests.

tests/batch_selection_test.py
vvv214: deleted

tests/batch_selection_test.py
vvv214: changed

vvv214: changed to .003

tests/batch_selection_test.py
vvv214: removed

vvv214:
thanks @arung54 for the review; i've made another commit (shouldn't 'force-push') addressing your comments

arung54 (approved):

PR #284, opened 2026-06-23
Validate structural arguments (num_bands, truncated sizes) in accounting events

r1cksync:
## Bug

Several `jax_privacy.accounting` event builders forward user-controlled structural arguments into divisions / sizing without validating them, so invalid inputs either crash opaquely or are silently accepted:

```python
>>> from jax_privacy import accounting
>>> accounting.amplified_bandmf_event(1.0, 128, num_bands=0, sampling_prob=0.01)
ZeroDivisionError: division by zero
>>> accounting.amplified_bandmf_event(1.0, 128, num_bands=-4, sampling_prob=0.01)
ValueError: Expected iterations=-32 >= 0.          # misleading: blames iterations
>>> accounting.truncated_dpsgd_event(1.0, 10, sampling_prob=0.1,
...     num_examples=-5, truncated_batch_size=-3)
<SelfComposedDpEvent ...>                           # silently accepted, nonsensical
```

`num_bands` flows into `math.ceil(iterations / num_bands)` in `amplified_bandmf_event` and `truncated_amplified_bandmf_event`; `num_examples` / `truncated_batch_size` flow into `TruncatedSubsampledGaussianDpEvent` in `truncated_dpsgd_event`. None were validated, while sibling builders (`fixed_dpsgd_event` via `_validate_fixed_args`) already validate their sizes.

## Fix

- `_validate.positive(num_bands=num_bands)` in both BandMF builders (must be ≥ 1 to partition the data and to divide by).
- `_validate.non_negative(num_examples=..., truncated_batch_size=...)` in `truncated_dpsgd_event`, matching the `_validate_fixed_args` convention for dataset/batch sizes.

Only previously-invalid inputs (non-positive `num_bands`, negative sizes) are rejected — valid inputs are unchanged. This mirrors the already-merged #255 and the existing `_validate` pattern.

## Testing

`python -m pytest tests/accounting_test.py` — 36 passed. The new parameterized tests cover each newly-guarded argument × invalid value, plus positive regressions; they fail (8 failures + 2 ZeroDivisionError) against the pre-fix code and pass after.

### DIFF
--- a/tests/accounting_test.py
+++ b/tests/accounting_test.py
@@ -233,6 +233,95 @@ class AccountingTest(parameterized.TestCase):
     # as the continuous Gaussian with the same sigma.
     self.assertAlmostEqual(eps_continuous, eps_discrete, places=4)
 
+  # -- Structural-argument validation ----------------------------------------
+  # num_bands flows into math.ceil(iterations / num_bands); the truncated sizes
+  # flow into TruncatedSubsampledGaussianDpEvent. These were previously
+  # unvalidated: num_bands=0 raised an opaque ZeroDivisionError, num_bands<0
+  # raised a misleading "iterations=..." error, and negative sizes were silently
+  # accepted and produced a nonsensical event.
+
+  @parameterized.parameters(0, -1, -16)
+  def test_amplified_bandmf_event_rejects_nonpositive_num_bands(
+      self, num_bands
+  ):
+    with self.assertRaisesRegex(ValueError, rf"num_bands={num_bands} > 0"):
+      accounting.amplified_bandmf_event(
+          1.0, 128, num_bands=num_bands, sampling_prob=0.01
+      )
+
+  @parameterized.parameters(0, -1, -16)
+  def test_truncated_amplified_bandmf_rejects_nonpositive_num_bands(
+      self, num_bands
+  ):
+    with self.assertRaisesRegex(ValueError, rf"num_bands={num_bands} > 0"):
+      accounting.truncated_amplified_bandmf_event(
+          1.0,
+          128,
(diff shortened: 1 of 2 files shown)


tests/accounting_test.py
```diff
@@ -233,6 +233,95 @@ def test_use_zcdp_matches_gaussian_privacy(self):
...
+      accounting.truncated_amplified_bandmf_event(
+          1.0,
+          128,
+          num_bands=16,
+          sampling_prob=0.01,
+          largest_group_size=-1,
+          truncated_batch_size=16,
+      )
+
+  def test_amplified_bandmf_event_valid_num_bands_unchanged(self):
```
arung54, line 296: I think '_unchanged' here and in the below test names is unnecessary and if anything adds confusion.

tests/accounting_test.py
```diff
@@ -233,6 +233,95 @@ def test_use_zcdp_matches_gaussian_privacy(self):
     # as the continuous Gaussian with the same sigma.
     self.assertAlmostEqual(eps_continuous, eps_discrete, places=4)
 
+  # -- Structural-argument validation ----------------------------------------
```
arung54, line 236: I think none of the comments introduced in this CL are necessary, and would suggest removing them all.

From the test names I think it is pretty easy to infer what each test is supposed to do, and the logic can be inferred from reading the file being tested. We generally prefer not to include comments if they only explain something that can be inferred by someone with knowledge of Python.

arung54 (commented):
Thanks for the PR. No issues with the changes in logic, but a few comments to keep the style consistent. Will be happy to approve once they're addressed.

PR #304, opened 2026-07-08
refactor(deps): make chex a test-only dependency

### LINKED ISSUE #303: Remove chex dependency from core library
We should make chex a test-only dependency (see pyproject.toml), replacing chex.ArrayTree -> optax.ArrayTree and chex.dataclass with @jax.tree_util.register_jax_dataclass along with jax.tree.static() annotatoins on static fields.

Neal006:
Closes #303.

Removes `chex` from the core library so it is only needed for tests, as requested in #303.

## Changes
- **`chex.ArrayTree` → `optax.ArrayTree`** across `clipping`, `keras_api`, `sharding_utils`, `training`, `discrete_gaussian`, `buffered_toeplitz`, `optimization`, `streaming_matrix`.
- **`chex.Numeric` / `chex.Array` → `jax.typing.ArrayLike`** (inputs) / **`jax.Array`** (outputs), following the [type-annotations JEP](https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html) that `buffered_toeplitz` already cites.
- **`@chex.dataclass` → `@jax.tree_util.register_dataclass` + `@dataclasses.dataclass`** on `TrainingState`, `ColumnNormalizedBanded`, and `BufferedToeplitz`. All fields are pytree *data* fields in each of these, so no `jax.tree.static()` annotations are needed and the prior flatten/unflatten/`tree_map` behavior is preserved.
- **`pyproject.toml`**: `chex` moved from `dependencies` to the `[test]` optional-dependencies group.
- **`docs/conf.py`**: `autodoc_type_aliases['ArrayTree']` updated to `optax.ArrayTree`.

## Notes
- `optax.ArrayTree` is a strict superset of `chex.ArrayTree` (it additionally permits Python scalar leaves), so the annotation change only widens the accepted types.
- Verified the three dataclass conversions preserve chex pytree semantics (leaf count/ordering, flatten/unflatten roundtrip, `tree_map`, frozen enforcement, and `jit` tracing of all fields) against the original `@chex.dataclass` definitions.
- Formatting checked with `pyink` (repo config).

### DIFF
--- a/jax_privacy/matrix_factorization/buffered_toeplitz.py
+++ b/jax_privacy/matrix_factorization/buffered_toeplitz.py
@@ -24,7 +24,6 @@ import functools
 from typing import Any, TypeAlias
 
 from absl import logging
-import chex
 import jax
 import jax.numpy as jnp
 import numpy as np
@@ -48,7 +47,7 @@ from . import toeplitz
 
 
 ThetaPairType: TypeAlias = tuple[jax.Array, jax.Array]
-ScalarFloat: TypeAlias = chex.Numeric | float
+ScalarFloat: TypeAlias = jax.typing.ArrayLike
 
 
 def _gpu_or_cpu_device() -> jax.Device:
@@ -157,7 +156,8 @@ class StreamingMatrixBuilder:
     )
 
 
-@chex.dataclass
+@jax.tree_util.register_dataclass
+@dataclasses.dataclass
 class BufferedToeplitz:
   """A lower-triangular Toeplitz C parameterized as a BLT.
 
@@ -832,7 +832,7 @@ class Parameterization:
     loss_fn: The loss function to optimize.
   """
 
-  params_from_blt: Callable[[BufferedToeplitz], chex.ArrayTree]
+  params_from_blt: Callable[[BufferedToeplitz], optax.ArrayTree]
   blt_and_inverse_from_params: Callable[
       [Any],
       tuple[BufferedToeplitz, BufferedToeplitz],
@@ -880,7 +880,7 @@ class Parameterization:
 
   def get_loss_fn(
       self, loss_fn: LossFn
-  ) -> Callable[[chex.ArrayTree], ScalarFloat]:
+  ) -> Callable[[optax.ArrayTree], ScalarFloat]:
     """Returns a loss function for the parameterization."""
     return lambda params: loss_fn.penalized_loss(
         *self.blt_and_inverse_from_params(params)
@@ -1146,7 +1146,7 @@ def optimize(
 
 
 def geometric_sum(
-    a: jax.Array, r: jax.Array, num: chex.Numeric = jnp.inf
+    a: jax.Array, r: jax.Array, num: jax.typing.ArrayLike = jnp.inf
 ) -> jax.Array:
   """Sum a + a*r + a*r**2 + ... + a*r**(num-1) (or limit if num=jnp.inf).
 
@@ -1288,7 +1288,9 @@ def blt_pair_from_theta_pair(
 
 @jax.jit
 @require_buf_decay_less_eq_one
-def sensitivity_squared(blt: BufferedToeplitz, n: chex.Numeric) -> float:
+def sensitivity_squared(
+    blt: BufferedToeplitz, n: jax.typing.ArrayLike
+) -> float:
   """Computes sensitivity**2 for a BLT strategy matrix C.
 
   See https://arxiv.org/pdf/2404.16706 Lemma 5.3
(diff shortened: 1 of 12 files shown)


jax_privacy/matrix_factorization/buffered_toeplitz.py
```diff
@@ -350,6 +350,8 @@ def canonicalize(self) -> 'BufferedToeplitz':
     """Returns a `BufferedToeplitz` with buf_decay in decreasing order."""
     self.validate()  # Make sure lengths match.
     idx = jnp.argsort(self.buf_decay)[::-1]
+    # pylint does not recognize jax.Array as subscriptable.
```
ryan112358: This is surprising to me. Why was this not a problem before this PR? (or was it)

ryan112358 (changes requested):

jax_privacy/matrix_factorization/buffered_toeplitz.py
Neal006: @ryan112358 
this was not a problem before and it is not a real problem now. 
The line it guards, self.buf_decay[idx] on line 356, is unchanged by this PR. Only pylint's view of it changed.
Before, @chex.dataclass was opaque to pylint's inference engine, so it read self.buf_decay as Any and never flagged the indexing. 
After I switched to @dataclasses.dataclass + register_dataclass, pylint now reads the buf_decay: jax.Array annotation and infers the real type. 
Its model of jax.Array exposes no __getitem__, so it wrongly reports unsubscriptable-object on line 356. 
The indexing runs fine; this is just a pylint blind spot for jax.Array. That is why I added the scoped disable on line 354. The same cause shows up once more, in the inverse method. 
Line 413 runs omega2 = -omega and line 422 runs omega[:, jnp.newaxis] / (evals - blt.buf_decay[:, jnp.newaxis]). 
pylint flags the negation and the indexing there for the same reason, so I wrapped just those lines with the disable and enable pair on lines 411 and 423.
If you would rather avoid inline disables, I can fold them into one documented entry in .pylintrc instead. Your call.

ryan112358: Since this is a false positive, I think your suggestion of folding them into .pylintrc with a comment about why they ar ebeing disabled would be better (as you suggested)

Neal006: @ryan112358 , I have updated the changes, please review it and let me know if there are any more changes to do.
It's been a great learning till now and thank you for the opportunity.
Open to contribute more!!

ryan112358 (approved):

jax_privacy/matrix_factorization/buffered_toeplitz.py
ryan112358: Sounds good, I've approved here but it needs a second internal approval which can take some time before.  In the meantime, I think you can link this to the issue and mark it as fixed (or wait until it lands).

Neal006:
Hi @ryan112358 , thanks again for the review and approval. 
I can see feedback/copybara is red with "Google internal checks FAILED," but I'm not familiar with what that check covers or what it needs from my side. 
Could you point me to what's failing?
One thing I noticed: this PR adds a new import optax to a few files (keras_api.py, sharding_utils.py, experimental/discrete_gaussian.py, matrix_factorization/streaming_matrix.py) and drops chex from the core. If the internal build needs its dependency list updated for that, just let me know and I'll make whatever change you need on the GitHub side. Happy to help however I can to get this landed.

ryan112358:
Yes there seems to be an opaque formatting error internally, I can't tell what the problem is but it's something we may be able to handle on our end during merge.  @RamSaw is assigned for the internal review, and can follow-up if necessary

Neal006:
Thanks😊

Neal006:
Hey!!! @ryan112358 . I was just checking on to get some updates on this PR?

jax_privacy/matrix_factorization/streaming_matrix.py
```diff
@@ -20,12 +20,12 @@
 import dataclasses
 from typing import Any, Generic, TypeAlias, TypeVar
 
-import chex
 import jax
 from jax import numpy as jnp
+import optax
```
RamSaw: please remove this empty line to make internal formatter happy and I will merge it

RamSaw (changes requested):

jax_privacy/matrix_factorization/streaming_matrix.py
Neal006: @RamSaw 
I have done the changes!
Let me know, if there's anything else, you want me to do.
Thanks!

RamSaw (approved):

Neal006:
Hi @RamSaw, just checking in for updates regarding PR merge.
Thanks!

PR #333, opened 2026-07-24
Require jax>=0.8.1

Neal006:
Per discussion below: `DPTrainer.train_step` uses the bare `@jax.jit(...)` decorator, which is only valid on `jax>=0.8.1`. On 0.7.x and 0.8.0, `jax.jit` requires `fun` positionally, so the decorator raises `TypeError: jit() missing 1 required positional argument: 'fun'` at import time, breaking `jax_privacy.keras_api` and the Keras example notebooks on current Colab (jax 0.7.2).

Following @ryan112358's preference (option 2), this bumps the `jax` and `jaxlib` minimums to `0.8.1` rather than changing the decorator. This also covers the sharding-in-types fixes introduced in 0.8.0 and the optax features that require a newer jax.

I confirmed the crossover by testing across versions: the bare form fails on 0.7.x and 0.8.0 and works on 0.8.1 through 0.11.0.

cc @RamSaw

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,8 +28,8 @@ requires-python = ">=3.11"
 dependencies = [
     "absl-py>=1.2.0",
     "dp_accounting @ git+https://github.com/google/differential-privacy.git#subdirectory=python/dp_accounting",
-    "jax>=0.7.1",
-    "jaxlib>=0.7.1",
+    "jax>=0.8.1",
+    "jaxlib>=0.8.1",
     "numpy>=2.0.0",
     "optax @ git+https://github.com/google-deepmind/optax.git",
     "scipy>=1.9.2",


ryan112358:
What version of jax would we need to bump to for this to work?  Maybe it would be better to just update pyproject.toml and inside colab install a newer version of jax.  There are certain things that were fixed by version 0.8.0 around sharding-in-types that could cause problems on older versions of jax.

Neal006:
Hi @ryan112358 ,
Good call to check. I binary-searched it: the bare @jax.jit(...) factory form fails on jax 0.7.x and 0.8.0, but works from jax 0.8.1 onward. jax re-added a default for the fun argument in 0.8.1, so the fun-less decorator form is valid again. I confirmed 0.8.1 through 0.11.0 all accept it.
So there are two clean ways to fix this and I'm happy with either:
1. Keep the functools.partial wrap in this PR. It works across the whole current range (jax>=0.7.1), so it fixes the import without raising the floor or needing a jax upgrade in Colab.
2. Drop the code change and bump the pyproject minimum to jax>=0.8.1 (and jaxlib), so the original code works as written. One thing to flag: the sharding-in-types fixes you mentioned are in 0.8.0, but the jit decorator specifically needs 0.8.1, so the pin would have to be 0.8.1 rather than 0.8.0.
If you want to bump the minimum jax for the sharding-in-types fixes regardless, option 2 keeps the decorator as you wrote it. If you'd rather not raise the floor, option 1 is the safer minimum. Let me know which you prefer and I'll update the PR.

ryan112358:
Lets go with option (2) - there are some optax features we use that require higher versions of Jax anyway, e.g.,

https://github.com/google-deepmind/optax/blob/main/optax/microbatching/_microbatching.py#L111

ryan112358 (approved):

Neal006:
Hi, @ryan112358  sorry to make you go through the long way but the two failing public checks (accounting_test::test_calibration and toeplitz_test::test_helper_stddev_at_max_bands) are unrelated to this PR. This branch only bumps the jax/jaxlib version floors, and both are tight-tolerance numerical asserts drifting from the git-HEAD dp_accounting/optax pins. They reproduce on #335, and #334 is already fixing the calibration side. The internal feedback/copybara and import/copybara checks (pinned deps) both pass here. 
Happy to rebase once #334 lands so the public run goes green too, but there's nothing to change in this PR.

PR #337, opened 2026-07-28
Add DpsgdConfig execution plan for Poisson and fixed-size DP-SGD

Ayush7614:
## Summary

* Add a single `DpsgdConfig` Tier-2 `DPExecutionPlan` config for vanilla DP-SGD, with `DpsgdBatchSelection` (`POISSON` / `FIXED`) routing batch selection and matching accounting events.
* Mirror the existing `BandMFConfig` `calibrate()` / `make()` API, including `expected_participations` (no public `num_examples` required for add/remove Poisson) and privacy-correctness tests.
* Validation helpers live in `_validate` (`not_none`, `instance_of`, `at_most`); expected batch size is exposed as `expected_batch_size`.

## AI usage disclosure

AI (Cursor) was used to draft and iterate on this PR under my direction. I specified the API shape (unify configs, `expected_participations`, enum routing, drop `replace` / `use_zcdp`, lean validation, remove the extra example), ran and checked the test suite locally (`pytest tests/execution_plan_test.py`), and reviewed the resulting diffs before each push. Several review rounds with @ryan112358 drove further edits; I verified each change against the comments and re-ran tests after updates.

## Test plan

* [x] `pytest tests/execution_plan_test.py` (21 passed)
* [x] CI on this PR


jax_privacy/execution_plan.py
```diff
@@ -481,6 +481,402 @@ def clipped_grad_transform(*args, **kwargs):
...
+  """Raises ValueError if noise_multiplier has not been set."""
+  if noise_multiplier is None:
+    raise ValueError(
+        'noise_multiplier is not set. Call calibrate() or provide'
+        ' noise_multiplier directly.'
+    )
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class DpsgdConfig:
```
ryan112358, line 520: Let's try to unify DpsgdConfig and FixedSizeDpsgdConfig under a single dataclass, where the batch selection / accounting is routed based on one of the dataclass fields

jax_privacy/execution_plan.py
```diff
@@ -481,6 +481,402 @@ def clipped_grad_transform(*args, **kwargs):
...
+    >>> config = DpsgdConfig(
+    ...   iterations=1000, expected_batch_size=256, num_examples=50000,
+    ...   noise_multiplier=1.0,
+    ... )
+
+  References: https://arxiv.org/abs/1607.00133
+
+  Attributes:
+    iterations: Number of training iterations / batch draws.
+    expected_batch_size: Expected Poisson batch size. Sampling probability is
```
ryan112358: Being able to specify this implies you know the number of examples, which rules out ADD_OR_REMOVE neighboring relation.  Let's parameterize this in terms of "expected_participations" like BandMFConfig

jax_privacy/execution_plan.py
```diff
@@ -481,6 +481,402 @@ def clipped_grad_transform(*args, **kwargs):
...
+      Treated as public information for accounting.
+    noise_multiplier: Ratio of noise stddev to query sensitivity. If unset,
+      call ``calibrate()``.
+    l2_clip_norm: Maximum L2 norm of per-example gradients.
+    rescale_to_unit_norm: Divide clipped gradients by ``l2_clip_norm``.
+    normalize_by: Divide the sum-of-clipped gradients by this value. Defaults to
+      ``expected_batch_size`` when ``None``.
+    truncated_batch_size: Optional truncation cap for Poisson sampling. When
+      set, accounting uses ``truncated_dpsgd_event``.
+    use_zcdp: If True, account with a ``ZCDpEvent`` (e.g. for discrete
```
ryan112358: Discrete gaussian is currently not fully compatable with this API or the noise_addition API, so let's drop that reference and this flag here.

ryan112358 (commented):
Nice addition, thanks!  Please address the comments and I'll take another look

Ayush7614:
Thanks for the review @ryan112358 — addressed in the latest commit:

1. **Unified** `DpsgdConfig` and `FixedSizeDpsgdConfig` into a single `DpsgdConfig`, routed by `batch_size` (unset = Poisson, set = fixed-size).
2. **Poisson path** now uses `expected_participations` like `BandMFConfig` (no `num_examples` required under add/remove). `num_examples` is only required for fixed-size or truncated Poisson.
3. **Removed** `use_zcdp` and discrete-Gaussian references.

Happy to adjust further if needed.

jax_privacy/execution_plan.py
```diff
@@ -481,6 +481,293 @@ def clipped_grad_transform(*args, **kwargs):
...
+  iterations: int
+  expected_participations: float
+  noise_multiplier: float | None = None
+  l2_clip_norm: float = 1.0
+  rescale_to_unit_norm: bool = True
+  normalize_by: float = 1.0
+  batch_size: int | None = None
+  num_examples: int | None = None
+  truncated_batch_size: int | None = None
+  replace: bool = False
```
ryan112358: Let's drop this kwarg

jax_privacy/execution_plan.py
```diff
@@ -481,6 +481,293 @@ def clipped_grad_transform(*args, **kwargs):
...
+      ``truncated_batch_size`` is set (dataset size is treated as public, which
+      rules out add/remove adjacency for those paths). Not needed for plain
+      Poisson sampling under add/remove adjacency.
+    truncated_batch_size: Optional truncation cap for Poisson sampling. Requires
+      ``num_examples``. Incompatible with ``batch_size``.
+    replace: Whether fixed-size batches sample with replacement. Ignored for
+      Poisson sampling.
+  """
+
+  iterations: int
```
ryan112358, line 568: In the future, we will want to support more than just Poisson / Fixed Batch.  There's also Random allocation - support is being added for that in dp-accounting now.  Whatever design we have here needs to be done with that future in mind

jax_privacy/execution_plan.py
```diff
@@ -481,6 +481,293 @@ def clipped_grad_transform(*args, **kwargs):
...
+  expected_participations: float
+  noise_multiplier: float | None = None
+  l2_clip_norm: float = 1.0
+  rescale_to_unit_norm: bool = True
+  normalize_by: float = 1.0
+  batch_size: int | None = None
```
ryan112358: This can be inferred from expected_participations and num_examples so is probably not needed

ryan112358 (commented):
Thx for the quick turnaround, please respond to each comment individually or mark it as resolved. (from prior review & this one)

jax_privacy/execution_plan.py
Ayush7614: Done — merged into a single `DpsgdConfig`. Routing is now via an explicit `batch_selection: DpsgdBatchSelection` field (`POISSON` / `FIXED`), so additional strategies can be added later without a second config class.

jax_privacy/execution_plan.py
Ayush7614: Done — Poisson path now takes `expected_participations` (same idea as `BandMFConfig`). `num_examples` is only required for fixed-size (and truncated Poisson) where the dataset size is needed to derive the batch size / sampling probability.

jax_privacy/execution_plan.py
Ayush7614: Done — removed `use_zcdp` and the discrete-Gaussian references from the config and docs.

jax_privacy/execution_plan.py
Ayush7614: Done — dropped the `replace` kwarg.

jax_privacy/execution_plan.py
Ayush7614: Agreed — switched the routing field to an explicit `DpsgdBatchSelection` enum (`POISSON`, `FIXED`) so we can extend with `RANDOM_ALLOCATION` (or others) once dp-accounting support lands, without another config class or a boolean/`batch_size`-as-sentinel design.

jax_privacy/execution_plan.py
Ayush7614: Done — removed `batch_size` as a constructor argument. For `FIXED`, it is now derived as `expected_participations * num_examples / iterations` (exposed as a read-only property for callers that still need it).

Ayush7614:
Thanks @ryan112358 — replied on each thread. Summary of the latest push:

- **`DpsgdBatchSelection` enum** routes Poisson vs fixed-size (extensible for random allocation later)
- Dropped **`replace`** and constructor **`batch_size`** (fixed batch size inferred from `expected_participations` + `num_examples`)
- Earlier feedback already applied: unified config, `expected_participations`, no `use_zcdp`

Happy to adjust further if the enum / inference shape isn’t quite right.

jax_privacy/execution_plan.py
```diff
@@ -481,6 +482,348 @@ def clipped_grad_transform(*args, **kwargs):
...
+        dtype=performance_flags.dtype,
+        microbatch_size=performance_flags.microbatch_size,
+        spmd_axis_name=performance_flags.spmd_axis_name,
+        keep_batch_dim=performance_flags.keep_batch_dim,
+    )
+
+  return clipped_grad_transform
+
+
+def _check_noise_multiplier_set(noise_multiplier: float | None) -> None:
```
ryan112358: Maybe add this to _validate.py as say "def notnull(**kwargs)" or similar

jax_privacy/execution_plan.py
```diff
@@ -481,6 +482,348 @@ def clipped_grad_transform(*args, **kwargs):
...
+    if self.noise_multiplier is not None:
+      _validate.non_negative(noise_multiplier=self.noise_multiplier)
+
+    if not isinstance(self.batch_selection, DpsgdBatchSelection):
+      raise TypeError(
+          'batch_selection must be a DpsgdBatchSelection, got'
+          f' {type(self.batch_selection)!r}.'
+      )
+
+    if (
```
ryan112358: This is adding a significant amount of bloat.  Please think carefully about if it can be reduced or otherwise factored into _validate.py

jax_privacy/execution_plan.py
```diff
@@ -481,6 +482,348 @@ def clipped_grad_transform(*args, **kwargs):
...
+            'truncated_batch_size requires num_examples to be set.'
+        )
+      if self.truncated_batch_size > self.num_examples:
+        raise ValueError(
+            f'truncated_batch_size={self.truncated_batch_size} must be <='
+            f' num_examples={self.num_examples}.'
+        )
+
+  @property
+  def batch_size(self) -> int:
```
ryan112358: For poisson we have variable batch sizes, so this would have to be 'expected_batch_size'

examples/dp_sgd_execution_plan.py
```diff
@@ -0,0 +1,138 @@
+# coding=utf-8
```
ryan112358: Generally we prefer smaller focused PRs and we pay close attention to "diff hygiene" and whether a change warrants the maintenance burden.  Given that users can (or should be able to) just take an existing example and swap out the execution plan, I think a totally separate example is unnecessary here, so I'd remove this here align with the change to examples_guide.md

ryan112358 (commented):
Please update the PR description to disclose how you used AI for this PR.  Things like how did you guide it, how did you check it's work, how much iteration did you do before you opened the PR, etc.  AI usage is generally fine, but reviewing has a cost so it's helpful to know

jax_privacy/execution_plan.py
Ayush7614: Done — added `_validate.not_none(**kwargs)` and use it from `DpsgdConfig.make()` (and BandMF’s calibrated check) instead of a one-off helper.

jax_privacy/execution_plan.py
Ayush7614: Done — pulled common checks into `_validate.not_none` / `instance_of` / `at_most` and slimmed `__post_init__` accordingly (net ~150 lines removed in this push with the example deletion).

jax_privacy/execution_plan.py
Ayush7614: Done — renamed the property to `expected_batch_size` (float). Fixed-size sampling still validates/uses an integer via a private `_fixed_batch_size()` helper.

examples/dp_sgd_execution_plan.py
Ayush7614: Done — removed `examples/dp_sgd_execution_plan.py` and the corresponding `examples_guide.md` / CHANGELOG references. Existing examples can swap in `DpsgdConfig` as needed.

Ayush7614:
Thanks @ryan112358 — addressed the latest review in `4d8c3df`:

- `_validate.not_none` / `instance_of` / `at_most`; slimmer `DpsgdConfig` validation
- renamed property to `expected_batch_size`
- removed the standalone example + docs/CHANGELOG mentions
- updated the PR description with an AI-usage note

Happy to adjust further if anything still feels heavy.

jax_privacy/execution_plan.py
```diff
@@ -481,6 +478,325 @@ def clipped_grad_transform(*args, **kwargs):
...
+        normalize_by=self.normalize_by,
+    )
+    _validate.in_range(
+        0,
+        self.iterations,
+        expected_participations=self.expected_participations,
+    )
+    if self.noise_multiplier is not None:
+      _validate.non_negative(noise_multiplier=self.noise_multiplier)
+    _validate.instance_of(
```
ryan112358: The type checker should catch this at "compile" time, no need for runtime validation

jax_privacy/execution_plan.py
```diff
@@ -481,6 +478,325 @@ def clipped_grad_transform(*args, **kwargs):
...
+
+    Equals ``expected_participations * num_examples / iterations``. Under
+    Poisson sampling this is the *expected* (variable) batch size; under
+    ``FIXED`` sampling it must be an integer and equals the fixed batch size.
+
+    Raises:
+      ValueError: If ``num_examples`` is unset.
+    """
+    _validate.not_none(num_examples=self.num_examples)
+    return self.expected_participations * self.num_examples / self.iterations
```
ryan112358, line 631: This might not be an integer under FIXED

jax_privacy/execution_plan.py
```diff
@@ -481,6 +478,325 @@ def clipped_grad_transform(*args, **kwargs):
...
+              if self.num_examples is not None
+              else batch_selection.PartitionType.INDEPENDENT
+          ),
+      )
+    else:
+      raise NotImplementedError(
+          f'Unsupported batch_selection={self.batch_selection!r}. Additional'
+          ' strategies will be wired as dp-accounting support lands.'
+      )
+    query_sensitivity = clipped_grad_transform(lambda: None).sensitivity(
```
ryan112358: So I think dp_accounting treats "noise_multiplier" as relative to "l2_norm_bound" which is 2x smaller than "sensitivity" under REPLACE_ONE DP.  It's worth changign this to l2_norm_bound and adding an inline comment explaining why, along with a test that noise get's calibrated correctly in this setting.

ryan112358 (changes requested):
Mostly looks good, will send for an internal review next.  Thanks!

jax_privacy/execution_plan.py
Ayush7614: Done — removed the runtime `instance_of` check (and the unused `_validate.instance_of` helper).

jax_privacy/execution_plan.py
Ayush7614: Done — clarified that `expected_batch_size` is always a float and need not be integral under `FIXED`. The integer passed to `FixedBatchSampling` is derived separately via `_fixed_batch_size()`, which still requires an exact positive integer for accounting consistency.

jax_privacy/execution_plan.py
Ayush7614: Done — `make()` now scales noise by `l2_norm_bound` (with an inline comment explaining why `sensitivity()` would over-noise under `REPLACE_ONE`). Added `test_fixed_size_noise_matches_l2_norm_bound_not_sensitivity` to lock this in.

Ayush7614:
Thanks @ryan112358 — addressed the latest review:

- Dropped runtime `instance_of` validation
- Clarified `expected_batch_size` may be non-integral under FIXED
- Noise now uses `l2_norm_bound` (not `sensitivity()`), with a REPLACE_ONE calibration test

`pytest tests/execution_plan_test.py`: 22 passed.

ryan112358 (approved):

jax_privacy/execution_plan.py
```diff
@@ -481,6 +478,328 @@ def clipped_grad_transform(*args, **kwargs):
...
+    ``FixedBatchSampling`` is derived separately via ``_fixed_batch_size``.
+
+    Raises:
+      ValueError: If ``num_examples`` is unset.
+    """
+    _validate.not_none(num_examples=self.num_examples)
+    return self.expected_participations * self.num_examples / self.iterations
+
+  def _fixed_batch_size(self) -> int:
+    """Integer batch size required by ``DpsgdBatchSelection.FIXED``.
```
arung54, line 634: Our linter is blocking approval since this multi-line docstring doesn't have a "Returns:"

jax_privacy/execution_plan.py
```diff
@@ -481,6 +478,328 @@ def clipped_grad_transform(*args, **kwargs):
...
+    """Returns a DpEvent for the configured DP-SGD variant."""
+    if self.batch_selection is DpsgdBatchSelection.FIXED:
+      return accounting.fixed_dpsgd_event(
+          noise_multiplier=sigma,
+          iterations=self.iterations,
+          dataset_size=self.num_examples,
+          batch_size=self._fixed_batch_size(),
+          replace=False,
+      )
+    if self.batch_selection is not DpsgdBatchSelection.POISSON:
```
arung54, line 672: Style nitpick: For readability/consistency with similar statement in other parts of this library/related libraries could you reformat this as:

```
if self.batch_selection is DpsgdBatchSelection.FIXED:
  ...
elif self.batch_selection is DpsgdBatchSelection.POISSON:
  ...
else:
  raise NotImplementedError(...)
```

jax_privacy/execution_plan.py
ryan112358: Probably for fixed_batch_size this should still be an integer (rounded) since by definition it has to be fixed across all steps.  I think you should fold in the logic from the _fixed_batch_size helper here, and delete that method.

jax_privacy/execution_plan.py
ryan112358: Agreed

jax_privacy/execution_plan.py
```diff
@@ -481,6 +478,328 @@ def clipped_grad_transform(*args, **kwargs):
...
+        (PLD does not currently support
+        ``SampledWithoutReplacementDpEvent``).
+
+    Returns:
+      A new ``DpsgdConfig`` with calibrated ``noise_multiplier``.
+    """
+    if accountant_fn is None:
+      if self.batch_selection is DpsgdBatchSelection.FIXED:
+
+        def _rdp_accountant(
```
ryan112358, line 716: Not sure what this does, can it be deleted?

jax_privacy/execution_plan.py
```diff
@@ -481,6 +478,328 @@ def clipped_grad_transform(*args, **kwargs):
...
+          iterations=self.iterations,
+          replace=False,
+      )
+    elif self.batch_selection is DpsgdBatchSelection.POISSON:
+      batch_selection_strategy = batch_selection.CyclicPoissonSampling(
+          sampling_prob=self.sampling_prob,
+          iterations=self.iterations,
+          cycle_length=1,
+          truncated_batch_size=self.truncated_batch_size,
+          partition_type=(
```
ryan112358, line 774: Since we have cycle_length=1, partition_type is a no-op and can be removed.

ryan112358:
Hi @Ayush7614 , I have mulled over this change a bit and have a few thoughts.

DpSGDConfig is a special case of BandMFConfig in most cases, with the exception of fixed batch sampling (which is not SOTA).  For that reason, the maintenance burden of this extra code is likely not worth the delta in capabilities it brings to Jax privacy **currently**.

With that said, if you are able to land a couple of other PRs before this one, then this class can feature some nice additional capabilities.  Specifically, if you can generalize DpSgd accounting to support a list of sampling probabilities and a list of batch sizes respectively, along with a list of noise multipliers, this would provide enough of a difference  to potentially warrant bringing this in.  

If you want to work on this, please open an issue first with your proposed change including what files it affects and how.

ryan112358:
>  Specifically, if you can generalize DpSgd accounting to support a list of sampling probabilities and a list of batch sizes respectively, along with a list of noise multipliers, this would provide enough of a difference to potentially warrant bringing this in.

@Ayush7614 is this something you'd be interested in contributing, or should I close this PR for now?

Issue #317, opened 2026-07-20
Design proposal: integrate SlaClip's private Slack Indicator and threshold controller


ZsyRock:
Following maintainer guidance, I am opening this issue for design discussion before starting an implementation.

### 1. Proposal

I would like to contribute a paper-faithful implementation of [SlaClip: Gradient Norm Slacks can be Indicator for Adaptive Clipping in DP-SGD](https://icml.cc/virtual/2026/poster/66390) (ICML 2026).

SlaClip consists of two composable components:

1. A **private Slack Indicator mechanism**, which is the main technical contribution of SlaClip.
   For each per-example gradient `g_i`, clipping threshold `C_t`, and slack dimension `K`, it constructs `g_i^+ = (clip_C_t(g_i), s_i)`, where `s_i` encodes the unused norm margin in `K` bounded coordinates.
   The paper proves `||g_i^+||_2 <= C_t`, so its fixed-denominator average has the same `C_t / B` add/remove sensitivity as the ordinary clipped-gradient average.
   The gradient and Slack Indicator can therefore be released as coordinates of one Gaussian mechanism under the usual DP-SGD accounting, yielding a noisy, binned estimate of the gradient-norm CDF on `[0, C_t]` while leaving the ordinary DP gradient coordinates unchanged in distribution.
2. A **threshold controller** that consumes the noised Slack Indicator and applies the published update for `C_{t+1}` as post-processing.

The private Slack Indicator can be used independently, while composing both components gives the published SlaClip method.

Raw per-example norms, per-example slack, the exact empirical CDF, and the unnoised slack aggregate remain private intermediates and would not be returned or logged.

### 2. Proposed JAX Privacy integration

I propose preserving the same two component boundaries in JAX Privacy.

For the private Slack Indicator, the gradient and slack can remain separate PyTree leaves for computation and aggregation while being jointly noised through the existing DP query abstractions, so no flat `d + K` tensor is required.

For the controller, `C_t` would be maintained in immutable JAX state and updated only after the current noised Slack Indicator has been released.

My initial preference is a self-contained experimental implementation limited to scalar global clipping, explicit static `K`, and i.i.d. Gaussian noise.

Tests would verify the joint norm bound, the ordinary gradient path, the published controller, JAX transformations, and unchanged privacy accounting.

If maintainers prefer a reusable core hook or integration with the existing trainer, I would be happy to adjust the design accordingly.

### 3. Questions for maintainers

[...]

PR #343, opened 2026-07-31
Add optional slack output to clipped gradients

ZsyRock:
## Summary

This PR is limited to the clipping primitive discussed in #317 and changes
only `jax_privacy/clipping.py` and `tests/gradient_clipping_test.py`.

- Add an optional positive integer `slack` argument to `clipped_fun` and
  `clipped_grad`.
- Build prefix-filled slack coordinates from the norm already computed by the
  existing clipping path, avoiding a second full-gradient norm traversal.
- Return `ClippedGradOutput(gradient, slack)` as one jointly
  sensitivity-bounded primary output. Both fields must be privatized together
  before slack is exposed.
- Preserve the original output structure and compiled path when `slack=None`.

## Design and scope

Slack construction is isolated in `_slack_from_norm`. It runs after
`pre_clipping_transform`, supports a dynamic scalar clipping bound, leaves the
gradient dtype unchanged, and emits float32 slack coordinates. The integer
`slack` argument remains static because it determines the output shape.

This first PR provides only the core private Slack Indicator primitive. It does
not add trainer integration, an adaptive threshold controller, CLI flags, or
CDF logging. A later controller can consume the privatized slack output without
changing this component.

The initial scope supports scalar global clipping. Discrete grid clipping and
PyTree clipping bounds are rejected when slack is enabled; per-leaf slack can
be added separately.

## Validation

- 2,245 tests passed across the repository's CI-compatible test shards.
- 16 module doctests passed.
- Pyink, Flake8, Pylint, Pytype, and Pyrefly passed.
- The lowered StableHLO for the disabled path is identical to upstream. The
  enabled path reuses the existing norm and adds only the slack construction
  and aggregation, not another full-gradient clipping pass.

### DIFF
--- a/jax_privacy/clipping.py
+++ b/jax_privacy/clipping.py
@@ -20,7 +20,7 @@ import dataclasses
 import functools
 import math
 import numbers
-from typing import Any, Callable, TypeAlias
+from typing import Any, Callable, NamedTuple, TypeAlias
 
 import dp_accounting
 import jax
@@ -35,6 +35,70 @@ ClippedTreeAndNorm = collections.namedtuple(
     '_ClippedTreeAndNorm', ['clipped', 'norm']
 )
 _REPLACE_SPECIAL = dp_accounting.NeighboringRelation.REPLACE_SPECIAL
+
+
+class ClippedGradOutput(NamedTuple):
+  """Gradient and slack fields forming a single primary query.
+
+  Both fields must be privatized together before ``slack`` is released.
+  """
+
+  gradient: PyTree
+  slack: jax.Array
+
+
+def _validate_slack_config(
+    slack_k: int | None,
+    l2_clip_norm: float | PyTree,
+    grid_scale: int | None,
+) -> float | None:
+  """Validates the clipping configurations supported by slack outputs."""
+  if slack_k is None:
+    return None
+  if isinstance(slack_k, bool) or not isinstance(slack_k, int):
+    raise TypeError('slack_k must be a Python integer or None.')
+  if slack_k <= 0:
+    raise ValueError('slack_k must be positive.')
+  if grid_scale is not None:
+    raise ValueError('slack_k is not supported with discrete grid clipping.')
+  if not jnp.isscalar(l2_clip_norm):
+    raise ValueError('slack_k requires static scalar global clipping.')
+  try:
+    clip_norm = float(l2_clip_norm)
+  except (TypeError, ValueError) as error:
+    raise TypeError('slack_k requires a static l2_clip_norm.') from error
+  if not math.isfinite(clip_norm) or clip_norm <= 0:
+    raise ValueError('slack_k requires a finite, positive l2_clip_norm.')
+  min_clip_norm = math.sqrt(float(jnp.finfo(jnp.float32).tiny))
+  max_clip_norm = math.sqrt(float(jnp.finfo(jnp.float32).max))
+  if not min_clip_norm <= clip_norm <= max_clip_norm:
+    raise ValueError(
+        'slack_k requires l2_clip_norm whose square is finite and normal '
+        'float32.'
+    )
+  return clip_norm
+
+
+def _slack_from_norm(
+    gradient_norm: jax.Array,
+    clipping_bound: float,
+    slack_k: int,
+    output_bound: float,
+) -> jax.Array:
+  """Constructs prefix-filled slack from an already-computed norm.
+
+  The squared slack fraction is at most the unused norm fraction, which keeps
+  the gradient and slack fields jointly bounded by ``output_bound``.
+  """
+  clip_norm = jnp.asarray(clipping_bound, dtype=gradient_norm.dtype)
+  k = jnp.asarray(slack_k, dtype=gradient_norm.dtype)
+  unused_fraction = jnp.clip((clip_norm - gradient_norm) / clip_norm, 0.0, 1.0)
+  fill = jnp.clip(
+      k * unused_fraction - jnp.arange(slack_k, dtype=gradient_norm.dtype),
+      0.0,
+      1.0,
+  )
+  return fill * jnp.asarray(output_bound, gradient_norm.dtype) / jnp.sqrt(k)
 
 
 @dataclasses.dataclass(frozen=True)
@@ -424,6 +488,7 @@ def clipped_fun(
     prng_argnum: int | None = None,
     spmd_axis_name: str | None = None,
     grid_scale: int | None = None,
+    slack_k: int | None = None,
 ) -> BoundedSensitivityCallable:
   """Transforms a function to clip its output and sum across a batch.
 
@@ -498,6 +563,11 @@ def clipped_fun(
       Gaussian mechanism.  Incompatible with ``rescale_to_unit_norm=True`` and
       ``normalize_by != 1.0``.  When set, ``dtype`` is ignored (output is always
       ``jnp.int64``).
+    slack_k: If set, adds this many SlaClip coordinates as a separate slack
+      field of the sensitivity-bounded primary output. The returned value is a
+      :class:`ClippedGradOutput` whose fields must be privatized together. Only
+      static scalar global clipping with float32-or-wider primary outputs is
+      currently supported.
 
   Returns:
     A `BoundedSensitivityCallable` wrapping a new function `clip_fn` that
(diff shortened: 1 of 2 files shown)


ZsyRock:
@ryan112358 This is now ready for review. It implements the single-pass clipped_grad integration discussed in #317, and all public CI checks pass.

ryan112358:
Thanks for the work on this @ZsyRock .  This is a pretty large change and well beyond what we'd be able to review as a single PR.  Are all changes necessary?  Can some be simplified?  Is there a natural way some of the changes can be split up?

jax_privacy/clipping.py
ryan112358, line 1: As a concrete suggestion, for this first PR I think you should focus on the changes to clipping.py and gradient_clipping_test.py only.  This will be a manageable scope to review.  In terms of code organization, it's ideal if you can create a standalone helper function that consumes the gradient norm, clipping bound, slack parameter, etc. and returns the slack vector.  Ideally, the internal changes you need to make to the existing function(s) in this file should be very minimal (check if slack param is set, if so call your helper function).  This should help with readability & maintainability.

I may have over-simplified things above, since I assume things like "rescale_to_unit_norm" and some of the other fields might require other changes, but hopefully some of the design ideas above still apply.

ZsyRock:
Updated as requested: this PR is now a single commit changing only `jax_privacy/clipping.py` and `tests/gradient_clipping_test.py`. The implementation isolates slack construction in `_slack_from_norm` and reuses the norm returned by the existing clipping path. Trainer integration, the adaptive controller, CLI flags, and logging have been removed from this PR and can be proposed separately.

jax_privacy/clipping.py
```diff
@@ -37,6 +37,70 @@
...
+  """Validates the clipping configurations supported by slack outputs."""
+  if slack_k is None:
+    return None
+  if isinstance(slack_k, bool) or not isinstance(slack_k, int):
+    raise TypeError('slack_k must be a Python integer or None.')
+  if slack_k <= 0:
+    raise ValueError('slack_k must be positive.')
+  if grid_scale is not None:
+    raise ValueError('slack_k is not supported with discrete grid clipping.')
+  if not jnp.isscalar(l2_clip_norm):
```
ryan112358, line 65: Theoretically you could return slack on a per-leaf basis if you wanted to, probably out of scope for this PR, but maybe you can leave a TODO to flag this

jax_privacy/clipping.py
```diff
@@ -37,6 +37,70 @@
...
+    raise ValueError('slack_k requires static scalar global clipping.')
+  try:
+    clip_norm = float(l2_clip_norm)
+  except (TypeError, ValueError) as error:
```
ryan112358: Why do we require a static l2_clip_norm for slack to work?  I thought part of the reason to use slack was to dynamically update the l2_clip_norm across training?

jax_privacy/clipping.py
```diff
@@ -37,6 +37,70 @@
...
+  return clip_norm
+
+
+def _slack_from_norm(
+    gradient_norm: jax.Array,
+    clipping_bound: float,
+    slack_k: int,
+    output_bound: float,
+) -> jax.Array:
+  """Constructs prefix-filled slack from an already-computed norm.
```
ryan112358: Our formatter might complain about this docstring.  Usually it has to be a 1-line docstring or a full docstring with Args / Returns section.  I usually lean towards the 1-line docstring for private helper functions, and put the additional info in inline comments

jax_privacy/clipping.py
```diff
@@ -424,6 +488,7 @@ def clipped_fun(
     prng_argnum: int | None = None,
     spmd_axis_name: str | None = None,
     grid_scale: int | None = None,
+    slack_k: int | None = None,
```
ryan112358: I think I prefer the name "slack" over "slack_k" as long as the docstring is clear it has to be a int

jax_privacy/clipping.py
```diff
@@ -550,9 +623,33 @@ def clipped_fun_one_group(*args, is_padding_example, **kwargs):
       value = optax.tree.cast(value, dtype)
       if grid_scale is not None:
         clipped, norm = clip_and_round_to_grid(value, l2_clip_norm, grid_scale)
+        safety_norm = norm
+      elif slack_k is not None:
```
ryan112358: Can more of this be folded into the helper functions by any chance?

jax_privacy/clipping.py
```diff
@@ -550,9 +623,33 @@ def clipped_fun_one_group(*args, is_padding_example, **kwargs):
...
+        if any(
+            not jnp.issubdtype(x.dtype, jnp.floating)
+            or jnp.finfo(x.dtype).bits < 32
+            for x in jax.tree.leaves(value)
+        ):
+          raise TypeError(
```
ryan112358: Do you need this? I think you can return the slack in float32, but that shouldn't require the per-example grad leaves to have that dtype

jax_privacy/clipping.py
```diff
@@ -550,9 +623,33 @@ def clipped_fun_one_group(*args, is_padding_example, **kwargs):
...
+              'slack_k requires float32 or wider primary outputs; use'
+              ' dtype=jnp.float32.'
+          )
+        assert slack_clip_norm is not None
+        clipped, norm = clip_pytree(
+            value, slack_clip_norm, rescale_to_unit_norm
+        )
+        output_bound = 1.0 if rescale_to_unit_norm else slack_clip_norm
+        clipped = ClippedGradOutput(
```
ryan112358: nit: try to fit this on two lines by writing "slack = ..." then "clipped = ClippedGradOutput(gradient, slack)"

ryan112358 (changes requested):
Looking good!

jax_privacy/clipping.py
ZsyRock: Yes. Validation and all slack construction details are now contained in helpers. The existing clipping branch only creates the slack output after reusing its computed norm.

jax_privacy/clipping.py
ZsyRock: Done. The PR remains limited to clipping.py and gradient_clipping_test.py. Slack construction is isolated in _slack_from_norm, and the existing clipping path only validates the option and conditionally calls the helper.

jax_privacy/clipping.py
ZsyRock: Removed the static-value requirement. A scalar clipping bound may now be dynamic under JIT; the test executes one compiled function with two different bounds.

jax_privacy/clipping.py
ZsyRock: Done. The helper result is assigned to slack_vector, then passed directly to ClippedGradOutput; the name slack is already the integer argument.

jax_privacy/clipping.py
ZsyRock: Removed the gradient dtype restriction. The original gradient dtype is preserved, while the slack output is float32; float16 and bfloat16 are covered by tests.

jax_privacy/clipping.py
ZsyRock: Added a TODO for per-leaf slack while keeping this PR limited to scalar global clipping.

jax_privacy/clipping.py
ZsyRock: Changed the private helper to a one-line docstring and kept the bound argument in an inline comment.

jax_privacy/clipping.py
ZsyRock: Renamed the public argument to slack and documented that it must be a positive Python integer.

ZsyRock:
All requested changes are addressed in af64fc1, and all public CI checks, including the heavy tests, pass. The import/copybara check is showing the generic migration error; could you please retry or inspect the internal import when convenient?

ryan112358 (approved):

ZsyRock:
Thanks again for the approval. I noticed that the PR is now marked stale and is behind `main`, while the Copybara checks are green. Is it waiting on internal submission, or would you like me to rebase it onto `main`? I am happy to update it if needed.

ryan112358 (approved):

ryan112358:
Thanks for the follow-up, and sorry I lost track of this.  The PR looks good to me but I want to get a second set of eyes on it given it does touch "privacy-critical" code, will try to get back to you by Wednesday if any further changes are needed.

ZsyRock:
Hi Ryan, just following up on the internal check. I can see that `feedback/copybara` is failing, while `import/copybara` and all public checks are passing. Since the internal details are not visible externally, could you let me know whether any changes are needed on my side, or whether this is something handled internally?

PR #357, opened 2026-08-13
Fix empty user selection batch shape

anxkhn:
## Summary

Keep empty `UserSelectionStrategy` batches two-dimensional and integer typed.

When no users were sampled, NumPy produced a `(0,)` float array instead of the documented `(0, examples_per_user)` index matrix. Empty batches now preserve the signed index dtype and expected rank, including for an empty dataset.

## Tests

- Covered empty Poisson samples
- Covered empty datasets
- Verified shape `(0, 2)` and signed integer dtype


anxkhn:
hey @ryan112358, could you please review this pr when you have a chance?

ryan112358 (approved):
Sorry for the delayed review, looks good!

PR #372, opened 2026-09-10
Fix Keras train_steps to count optimizer updates

### LINKED ISSUE #234: Inconsistent `calculated_noise_multiplier` when using `gradient_accumulation_steps` in Gemma3 LoRA example

### LINKED ISSUE #288: Keras DP-SGD: `train_steps` disagrees by `gradient_accumulation_steps` between the accountant (optimizer updates) and the fit() budget check/docstring (batches)
## Summary

`DPKerasConfig.train_steps` is counted two different ways once `gradient_accumulation_steps` (G) `> 1`:

- The **privacy accountant** treats it as the number of **optimizer updates**: `dpsgd_event(iterations=train_steps, sampling_prob=effective_batch_size/train_size)` composes `train_steps` Poisson+Gaussian rounds (`keras_api.py:176-181` calibration, `:230-235` validation → `accounting.py:110-139`).
- The **fit()-time budget check and the docstring** treat it as the number of **size-`batch_size` batches**: `_get_default_steps_per_epoch(train_size, batch_size) = floor(train_size / batch_size)` (`keras_api.py:1081-1097`), and the docstring suggests `train_steps = epochs * (train_size // batch_size)` (`keras_api.py:101-105`).

With gradient accumulation the optimizer performs one update per `G` batches, so `floor(train_size / batch_size)` batches = `G ×` the optimizer updates. The two interpretations disagree by exactly `G`; only one can be correct.

## Why "optimizer updates" is the accounting-correct unit

The accountant composes one Poisson+Gaussian round per `train_steps`. Each round is one noisy weight release = one optimizer update, so `train_steps` must be the number of optimizer updates for the composition count to be right.

## Worked example

`train_size=50000, batch_size=256, gradient_accumulation_steps=4, epochs=20`:

- batches/epoch = `floor(50000/256) = 195`; total batches = `195 * 20 = 3900`.
- true optimizer updates = `3900 / 4 = 975`.
- the fit() budget check counts `3900` against `train_steps`; the docstring formula likewise gives `train_steps = 3900`.

So a user following the docstring sets `train_steps = 3900`, but the run performs only `975` updates, and the accountant composes `3900` rounds for a `975`-update run — `4×` too many. (The direction of *this* counting error is conservative — over-composing → over-noising and a wrong max-step limit — not a privacy under-count.)

## Code locations

- `keras_api.py:176-181` — calibration: `iterations = train_steps` (= optimizer updates).
- `keras_api.py:1081-1097` — `_get_default_steps_per_epoch` uses `batch_size`, no `/G`.
- `keras_api.py:701-717` — fit() budget guard compares batch counts against `train_steps`.
- `keras_api.py:101-105` — docstring says "optimizer update steps" but the formula counts size-`batch_size` batches.

## Proposed fix

1. Settle `train_steps` = **optimizer updates** (matching the accountant).
2. Divide the batch count by `gradient_accumulation_steps` in `_get_default_steps_per_epoch` / `_calculate_optimizer_steps_to_perform_in_fit`.
3. Fix the docstring formula to `epochs * (train_size // effective_batch_size)`.
4. Add a `gas > 1` test asserting the budget trips at the correct number of optimizer updates.

No existing test exercises `gradient_accumulation_steps > 1`, so this path is currently uncovered.

Anu27n:
## Summary
- The Gemma Keras notebooks counted `train_steps` as micro-batches (`epochs * (train_size // batch_size)`). With `gradient_accumulation_steps > 1` that over-composes the accountant and is why two runs with the same effective batch size produced different noise multipliers in #234.
- Examples and Keras API docs now set `train_steps = epochs * (train_size // effective_batch_size)`, matching `DPKerasConfig` and the accountant.
- `fit()` privacy-budget accounting now converts Keras micro-batch `train_step` counts into optimizer updates, so the budget check agrees with `train_steps` and the accountant (#288).

Related: #371 also touches Poisson sampling with gradient accumulation. This PR is scoped to example `train_steps` units and the `fit()` budget check.

Fixes #234.
Fixes #288.

## Test plan
- [ ] `KERAS_BACKEND=jax pytest tests/keras_api_test.py -k "gas or calibrated_noise_multiplier_invariant"`
- [ ] Confirm both Gemma notebooks pass optimizer-update `train_steps` into `DPKerasConfig`
- [ ] Confirm `fit()` with `gradient_accumulation_steps > 1` accepts in-budget optimizer updates and rejects over-budget ones


sylvesterkaczmarek (commented):
The update count is still wrong when the number of micro-batches per epoch is not divisible by `gradient_accumulation_steps`. Keras carries the accumulation state across epoch boundaries, but this helper rounds within each epoch and then multiplies by `epochs_to_perform`. For example, `train_size=100`, `batch_size=10`, `gas=4`, `epochs=3` gives 30 micro-batches and 7 optimizer updates (at global steps 4, 8, ..., 28), while this returns `floor(100/40) * 3 = 6`. The same remainder issue affects `ceil(performed_microbatch_steps / gas)`: after one micro-batch, zero optimizer updates have actually occurred, but it reports one. Could the budget calculation derive updates from the global micro-batch count/current accumulation remainder, with a regression where `steps_per_epoch % gas != 0`?

sylvesterkaczmarek (approved):
Re-reviewed current `ab02bd2e`. The accumulation remainder issue I raised is fixed: Keras progress now tracks microbatches globally and derives completed optimizer updates from the global count, so remainders carry across profile steps instead of being discarded per step. The new 15+15 microbatch regression with `grad_acc_steps=4` reaches exactly 7 optimizer iterations and pins the failure case. No remaining blocker from my review.

PR #392, opened 2026-09-17
Reject BandMFConfig.normalize_by <= 0 at construction

Anu27n:
## Summary
- `BandMFConfig` validated `normalize_by` as non-negative, so `0` constructed and then crashed in `clipped_grad`.
- Require `normalize_by > 0` at config construction, matching clipping.

## Test plan
- [ ] `pytest tests/execution_plan_test.py -k test_bandmf_validation`
- [ ] Confirm `BandMFConfig(..., normalize_by=0)` raises `ValueError`


sylvesterkaczmarek (commented):
`normalize_by=float("nan")` still passes the new validation. The shared `_validate.positive` helper only checks `value <= 0`, which is false for NaN, so a non-finite normalizer is accepted and can propagate NaNs into the normalized gradients/sensitivity instead of failing at construction. Could this require a finite positive value and add a NaN regression?

Anu27n:
Good catch — `_validate.positive` only checks `value <= 0`, so `NaN` slipped through.

Construction now requires `normalize_by` to be finite and `> 0` (`NaN` and `inf` both fail), with a NaN regression in `test_bandmf_rejects_nan_normalize_by`.

sylvesterkaczmarek (approved):
Rechecked 2bcd52f6. normalize_by now requires both finiteness and a positive value, so NaN and infinity are rejected at construction, with explicit regressions for both. This resolves the validation hole I raised. No blocker from me.

PR #394, opened 2026-09-17
Reject DPKerasConfig.delta values greater than 1

### LINKED ISSUE #384: DPKerasConfig accepts delta > 1 despite documenting (0, 1]
`DPKerasConfig` says `delta` must be in `(0, 1]`. Validation only uses `_validate.positive`, so `delta=0` is rejected but `delta=2.0` is accepted.

`tests/keras_api_test.py` covers `delta=0` only.

Anu27n:
## Summary
- `DPKerasConfig` documents `delta` in `(0, 1]` but only rejected `delta <= 0`.
- Reject `delta > 1` as well.

Fixes #384.

## Test plan
- [ ] `pytest tests/keras_api_test.py -k test_validate_params`
- [ ] Confirm `delta=2.0` raises `ValueError`


sylvesterkaczmarek (commented):
The new upper-bound check still leaves `NaN` accepted as a valid delta. `_validate.positive()` only tests `value <= 0` and this adds `value > 1`; both comparisons are false for `float('nan')`, so `DPKerasConfig(delta=nan, ...)` still gets past validation even though the documented domain is `(0, 1]`. Please validate the interval as a single positive finite condition (for example `not (0 < delta <= 1)`) and add a NaN regression before passing the value to the accountant.

Anu27n:
Agreed. `positive()` plus `delta > 1` both treat `NaN` as fine because those comparisons are false.

Delta is now checked with a single `0 < delta <= 1` condition, so `NaN`, `inf`, `0`, and values `> 1` all fail before the accountant runs. Added a NaN regression next to the existing `0` / `2.0` cases.

sylvesterkaczmarek (approved):
Rechecked c79340db. Delta is now validated with the single 0 < delta <= 1 condition, which rejects NaN as well as zero, infinity, and values above one before the accountant runs. The new NaN regression covers the case I raised. No blocker from me.

